PR #7, opened 2022-05-10
Agent nomination feature

mikedotexe:
This PR fleshes out the "agent nomination" process. When a task is added, sometimes the new number of tasks exceed the task ratio and a new agent can join. The agent who wishes to join has already registered as an agent, and is placed in the pending queue.

The moment that task (discussed previously) is added, a timestamp is set for when agent nominations are open.
There is a "duration" given for the first person in line. Once that duration has passed, the second person in the queue is allowed to "nominate themself" and be added as an active agent, and removed as a pending agent.

If the second person in line succeeds at nominating themself, everyone before them in line are removed from the pending queue, as they seem to be offline. They will need to re-register and get back in line.

---

We're using [`cw-multi-test`](https://github.com/CosmWasm/cw-plus/tree/main/packages/multi-test) to run the tests here, and in particular, work went into the test `accept_nomination_agent`. This will add the initial agent, give them tasks. Add two agents to pending queue. Add tasks until another agent is needed. "Fast forward" time only a few seconds such that only the first person in the pending queue is allowed to nominate themselves into being active. Show that the second person fails when trying to join, but the first succeeds. Then show the second person is now in the first position. Then add more tasks until another agent is needed again. Add another agent to the pending queue. Fast forward time such that the second person in the queue is allowed to nominate themselves. When this happens, confirm that the first person in line is essentially ejected from the queue and the second person is added as an active agent.

---

Added `AgentResponse` struct that contains the `status`, and removed it from the `Agent` struct since that doesn't need to be stored in state, it is only derived on a query.

---

Also removed the profiles section from `cw-croncat`'s manifest, as we kept getting:
```sh
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/mike/Documents/cw-croncat/contracts/cw-croncat/Cargo.toml
workspace: /Users/mike/Documents/cw-croncat/Cargo.toml
```
when building.


mikedotexe:
There were a few items Trevor and I discussed on a call today, and these are not yet added but will be. I wanted to give the opportunity for a head start. 

Items:

- [ ] agent_nomination_duration be optional with default value
- [ ] agent state should not change during accept_nomination_agent, just use queues
- [ ] check ContractError::CustomError and replace with UnregisterSomething instead

contracts/cw-croncat/src/agent.rs
```diff
@@ -671,15 +896,131 @@ mod tests {
...
+        let (agent_ids_res, num_active_agents, _) = get_agent_ids(&app, &contract_addr);
+        assert_eq!(1, num_active_agents);
+        assert_eq!(2, agent_ids_res.pending.len());
+
+        // Add two more tasks, so we can nominate another agent
+        add_task_exec(&mut app, &contract_addr, PARTICIPANT1);
+        add_task_exec(&mut app, &contract_addr, PARTICIPANT2);
+
+        num_tasks = get_task_total(&app, &contract_addr);
+        assert_eq!(num_tasks, 3);
```
TrevorJTClarke, line 941: I believe this should be 4.
Thoughts on this behaviour: 
ratio of 1:2, means entry upon >= 1:2 - so there is 1 agent doing up to 3 tasks before onboarding new agent. The task total would reach 4 tasks before adding another agent.
My main thinking here was to keep the agent profitable as a preference. More tasks is better for agents

contracts/cw-croncat/src/agent.rs
```diff
@@ -671,15 +896,131 @@ mod tests {
...
+        agent_status = get_agent_status(&mut app, &contract_addr, AGENT3);
+        assert_eq!(AgentStatus::Pending, agent_status);
+
+        // Again, add two more tasks so we can nominate another agent
+        add_task_exec(&mut app, &contract_addr, PARTICIPANT3);
+
+        add_task_exec(&mut app, &contract_addr, PARTICIPANT4);
+
+        // Add another agent, since there's now the need
+        register_agent_exec(&mut app, &contract_addr, AGENT4, &AGENT_BENEFICIARY);
```
TrevorJTClarke, line 1000: @mikedotexe this brings up a good point - if an agent registers and there are more than enough tasks to support them becoming active - do they skip the pending queue? Do they get put into the pending queue if pending queue is empty?

contracts/cw-croncat/src/tasks.rs
```diff
@@ -372,6 +372,27 @@ impl<'a> CwCroncat<'a> {
         // Add the attached balance into available_balance
         let mut c: Config = self.config.load(deps.storage)?;
         c.available_balance.add_tokens(Balance::from(info.funds));
+
+        // If the creation of this task means we'd like another agent, update config
+        let task_ratio = c.agent_task_ratio;
+        // Get total tasks
+        let total_tasks = self
```
TrevorJTClarke, line 379: @mikedotexe see other comments about the https://github.com/CronCats/cw-croncat/blob/main/contracts/cw-croncat/src/state.rs#L100 field & method - this should be incremented in this fn if it isnt already

TrevorJTClarke (changes requested):
@mikedotexe lots of great work here! See comments inline :) YOU ROCK!!!!

packages/cw-croncat-core/src/types.rs
```diff
@@ -51,6 +49,17 @@ pub struct Agent {
     pub register_start: u64,
```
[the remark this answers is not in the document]
mikedotexe: sure thing, just pushed it up

contracts/cw-croncat/src/agent.rs
```diff
@@ -2,40 +2,92 @@ use crate::error::ContractError;
...
+                    AgentStatus::Nominated
+                } else {
+                    AgentStatus::Pending
+                }
             } else {
-                a.status
+                // Not their time yet
+                AgentStatus::Pending
             }
+        } else if active.contains(&account_id) {
```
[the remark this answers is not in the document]
mikedotexe: good thought, just pushed up a commit addressing this https://github.com/CronCats/cw-croncat/pull/7/commits/eab3a31ad54278c00fdd73639d5ad71aa8ea942b

contracts/cw-croncat/src/agent.rs
```diff
@@ -2,40 +2,92 @@ use crate::error::ContractError;
...
+            AgentStatus::Active
         } else {
-            a.status
+            // This should not happen. It means the address is in self.agents
+            // but not in the pending or active queues
+            // Note: if your IDE highlights the below as problematic, you can ignore
+            return Err(StdError::GenericErr {
```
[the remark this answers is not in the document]
mikedotexe: Gotcha. The note about the IDE is because I'm using CLion and this is the first time I've experienced the IDE having red-underlines around this. It threw me for a loop trying to fix something that wasn't wrong. Thought I'd point it out for future devs who may see the same thing.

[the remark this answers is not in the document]
mikedotexe: Added the error in this commit https://github.com/CronCats/cw-croncat/pull/7/commits/6443161259793f87fb98aa5d0fc5a7585859ab89

contracts/cw-croncat/src/agent.rs
```diff
@@ -2,40 +2,92 @@ use crate::error::ContractError;
...
+            // Load config's task ratio, total tasks, active agents, and agent_nomination_begin_time.
+            // Then determine if this agent is considered "Nominated" and should call CheckInAgent
+            let agent_position = pending
+                .iter()
+                .position(|address| address == &account_id)
+                .unwrap();
+            let c: Config = self.config.load(deps.storage)?;
+            let task_ratio = c.agent_task_ratio;
+            // Get total tasks
+            let total_tasks = self
```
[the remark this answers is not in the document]
mikedotexe: Good idea, pushed up change in https://github.com/CronCats/cw-croncat/pull/7/commits/7b5c3fb102276c756d3eb9afd0a5ab602307e376

contracts/cw-croncat/src/agent.rs
```diff
@@ -2,40 +2,92 @@ use crate::error::ContractError;
...
+                .tasks
+                .keys(deps.storage, None, None, cosmwasm_std::Order::Ascending)
+                .count();
+            let num_active_agents = self
+                .agent_active_queue
+                .may_load(deps.storage)?
+                .unwrap_or_default()
+                .len();
+            // If we should allow a new agent to take over
+            if total_tasks as u64 > num_active_agents as u64 * task_ratio[0] * task_ratio[1]
```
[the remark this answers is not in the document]
mikedotexe: Split this into a helper and added a test for it here: https://github.com/CronCats/cw-croncat/pull/7/commits/1c06385708d81830a69381f4f9a34ea7524c2f6b

It was interesting to not use `app` in the tests at all for this one.

contracts/cw-croncat/src/agent.rs
```diff
@@ -249,10 +297,73 @@ impl<'a> CwCroncat<'a> {
...
+                c.agent_nomination_begin_time = None;
+                self.config.save(deps.storage, &c)?;
+            } else {
+                return Err(ContractError::CustomError {
+                    val: "Must wait longer before accepting nomination".to_string(),
+                });
+            }
+        } else {
+            // Sender's address does not exist in the agent pending queue
+            return Err(ContractError::Unauthorized {});
```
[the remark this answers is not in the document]
mikedotexe: Good call, changed here: https://github.com/CronCats/cw-croncat/pull/7/commits/4895ac516eec483e82aa4116d8c792ba03c2f19e

contracts/cw-croncat/src/agent.rs
mikedotexe: Interesting, thanks for talking that through on Element. As we discussed, I've changed the task ratio to be a single number that represents the maximum tasks an agent can do.
See: https://github.com/CronCats/cw-croncat/pull/7/commits/8fe414dd0d2d2cface528f749c2f6e256aa9c141

contracts/cw-croncat/src/tasks.rs
mikedotexe: Good eye. I just realized I called that earlier in the function so am reusing it. Saw it with fresh eyes today.
See https://github.com/CronCats/cw-croncat/pull/7/commits/6557ad98f00d04a9c2232b3bb982f38d8c53bba3

contracts/cw-croncat/src/state.rs
```diff
@@ -14,14 +14,24 @@ pub struct Config {
...
-    // The ratio of tasks to agents, where index 0 is agents, index 1 is tasks
-    // Example: [1, 10]
+    // The maximum number of tasks per agent
+    // Example: 10
     // Explanation: For every 1 agent, 10 tasks per slot are available.
-    // NOTE: Caveat, when there are odd number of tasks or agents, the overflow will be available to first-come first-serve. This doesnt negate the possibility of a failed txn from race case choosing winner inside a block.
+    // NOTE: Caveat, when there are odd number of tasks or agents, the overflow will be available to first-come first-serve. This doesn't negate the possibility of a failed txn from race case choosing winner inside a block.
     // NOTE: The overflow will be adjusted to be handled by sweeper in next implementation.
-    pub agent_task_ratio: [u64; 2],
+    pub max_tasks_per_agent: u64,
```
TrevorJTClarke, line 22: @mikedotexe i think it should actually be `min_tasks_per_agent`, because thats the minimum of new tasks needed before onboarding another agent. Thoughts?

contracts/cw-croncat/src/tasks.rs
```diff
@@ -340,8 +340,9 @@ impl<'a> CwCroncat<'a> {
             })?;
 
         // Increment task totals
-        let size: u64 = self.task_total.may_load(deps.storage)?.unwrap_or(0);
-        self.task_total.save(deps.storage, &(size + 1))?;
+        let mut size: u64 = self.task_total.may_load(deps.storage)?.unwrap_or(0);
+        size += 1;
```
TrevorJTClarke, line 344: @mikedotexe let's change to reuse this: https://github.com/CronCats/cw-croncat/blob/main/contracts/cw-croncat/src/state.rs#L104

contracts/cw-croncat/src/helpers.rs
```diff
@@ -52,6 +58,98 @@ pub(crate) fn has_cw_coins(coins: &[Cw20CoinVerified], required: &Cw20CoinVerifi
...
+        } else {
+            // This should not happen. It means the address is in self.agents
+            // but not in the pending or active queues
+            // Note: if your IDE highlights the below as problematic, you can ignore
+            return Err(AgentUnregistered {});
+        };
+        Ok(agent_status)
+    }
+
+    pub fn agents_to_let_in(
```
TrevorJTClarke, line 130: luvvit!

contracts/cw-croncat/src/helpers.rs
```diff
@@ -52,6 +58,98 @@ pub(crate) fn has_cw_coins(coins: &[Cw20CoinVerified], required: &Cw20CoinVerifi
         .unwrap_or(false)
 }
 
+impl<'a> CwCroncat<'a> {
+    pub fn get_agent_status(
+        &self,
+        storage: &dyn Storage,
+        // block_time: u64,
```
TrevorJTClarke, line 65: remove?

contracts/cw-croncat/src/helpers.rs
```diff
@@ -52,6 +58,98 @@ pub(crate) fn has_cw_coins(coins: &[Cw20CoinVerified], required: &Cw20CoinVerifi
...
+        env: Env,
+        account_id: Addr,
+    ) -> Result<AgentStatus, ContractError> {
```
TrevorJTClarke, line 68: nice work!

contracts/cw-croncat/src/agent.rs
```diff
@@ -148,13 +156,12 @@ impl<'a> CwCroncat<'a> {
...
                     None => {
                         Ok(Agent {
-                            status: agent_status.clone(),
                             payable_account_id: payable_id,
                             balance: GenericBalance::default(),
                             total_tasks_executed: 0,
                             last_missed_slot: 0,
                             // REF: https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/types.rs#L57
-                            register_start: env.block.time.nanos(),
+                            register_start: env.block.time,
```
TrevorJTClarke, line 164: Does this need to be `.seconds()`?

TrevorJTClarke (approved):
@mikedotexe throwing an approve on this - minor couple things in there if you have a minute :) Otherwise you rocked it!! :D

contracts/cw-croncat/src/state.rs
mikedotexe: Yeah, I agree. I suppose in my mind I thought it was "max tasks per agent for the system to allow someone in" but I've crammed more context in there than is clear from that name. Fixed here:
https://github.com/CronCats/cw-croncat/pull/7/commits/7270600b8f991066d35c8e10de13919fd96b5119

contracts/cw-croncat/src/tasks.rs
mikedotexe: Good stuff.
https://github.com/CronCats/cw-croncat/pull/7/commits/cadd6f4b9e9fd663f35bd0f605cc8e679fd41aa8

contracts/cw-croncat/src/helpers.rs
mikedotexe: Ah yes:
https://github.com/CronCats/cw-croncat/pull/7/commits/10179489334f998383e77d7cd9bfbac0e4ddbcb8

contracts/cw-croncat/src/agent.rs
mikedotexe: So this type was changed to `Timestamp` actually here:
https://github.com/CronCats/cw-croncat/blob/10179489334f998383e77d7cd9bfbac0e4ddbcb8/packages/cw-croncat-core/src/types.rs#L45-L49

```rs
pub struct Agent {
    …

    // Timestamp of when agent first registered
    // Useful for rewarding agents for their patience while they are pending and operating service
    // Agent will be responsible to constantly monitor when it is their turn to join in active agent set (done as part of agent code loops)
    // Example data: 1633890060000000000 or 0
    pub register_start: Timestamp,
}
```

Would you prefer it be a `u64`?

TrevorJTClarke: Ah i see - the context was lost in the implementation in review. Keep it, thanks

PR #12, opened 2022-06-23
Tests for get_current_slot_items, pop_slot_item

### LINKED ISSUE #9: [Tests] get_current_slot_items, pop_slot_item
There are two functions that need tests in `slots.rs`

In Croncat there is the idea of "slots" that are either based on time or block height. The time-based slot is equal to one minute.

## `get_current_slot_items`

This will return all the items in the current slot as well as any items that may not have been finished from previous slots.

## `pop_slot_item`

This removes the first item in the `Vec` of either the block or time-based slot. (See the `time_slots` and `block_slots` fields on `CwCroncat`.

I believe a good test to look at (and base the new tests from) is the `update_settings` test in `owner.rs`.

- [ ] `get_current_slot_items`
- [ ] `pop_slot_item`

Buckram123:
Closes #9 
Covered tests for `get_current_slot_items()` and `pop_slot_item()`
Optimized a bit `get_current_slot_items()`

### DIFF
--- a/contracts/cw-croncat/src/slots.rs
+++ b/contracts/cw-croncat/src/slots.rs
@@ -156,7 +156,6 @@ impl IntervalExt for Interval {
 }
 
 impl<'a> CwCroncat<'a> {
-    // TODO: TestCov
     /// Get the slot with lowest height/timestamp
     /// Returns (block slot, time slot)
     /// NOTE: This prioritizes blocks over timestamps.
@@ -194,7 +193,6 @@ impl<'a> CwCroncat<'a> {
         None
     }
 
-    // TODO: TestCov
     /// Gets 1 slot hash item, and removes the hash from storage
     /// Cleans up a slot if empty
     pub(crate) fn pop_slot_item(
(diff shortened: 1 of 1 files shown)


Buckram123:
Also wanted to ask if we should keep `mut` for `self` here? https://github.com/CronCats/cw-croncat/blob/c63bd6268ba6e8e238c5439c30b3b749246e8256/contracts/cw-croncat/src/slots.rs#L201-L202
It is enough for me to have `storage` as `mut` to visually understand that this method can change the state. Want to know your opinions.

One benefit is to avoid `clone()` here:
https://github.com/CronCats/cw-croncat/blob/c63bd6268ba6e8e238c5439c30b3b749246e8256/contracts/cw-croncat/src/slots.rs#L207-L208

contracts/cw-croncat/src/slots.rs
```diff
@@ -172,9 +171,9 @@ impl<'a> CwCroncat<'a> {
             .take(1)
             .collect();
 
-        if let Ok(block_id) = block_slot {
-            if !block_id.is_empty() && block_height >= block_id[0] {
-                return Some((block_id[0], SlotType::Block));
+        if let Ok(Some(block_id)) = block_slot.map(|v| v.first().copied()) {
```
mikedotexe, line 174: I like this, thank you

mikedotexe (approved):
very good work, I learned something about writing tests from this

PR #14, opened 2022-06-24
redundant may_load()'s

Buckram123:
Added initializing of `CwCroncat` `Item`'s  inside `instantiate()` method so we can get rid of 
`may_load().unwrap_or_default()` for them inside other methods


contracts/cw-croncat/src/agent.rs
```diff
@@ -24,10 +24,7 @@ impl<'a> CwCroncat<'a> {
         if agent.is_none() {
             return Ok(None);
         }
-        let active: Vec<Addr> = self
-            .agent_active_queue
-            .may_load(deps.storage)?
-            .unwrap_or_default();
+        let active: Vec<Addr> = self.agent_active_queue.load(deps.storage)?;
```
mikedotexe, line 27: Interesting, so this returns an empty Vec? For some reason I thought we needed the `.unwrap_or_default` in order to start with a default (empty) one

mikedotexe: I'm looking at the DAO DAO contracts, tho, and think you're right.
https://github.com/DA0-DA0/dao-contracts/blob/a531fe1a21d1c14c8f8461c94b0548e0e1f2293a/contracts/cw4-voting/src/contract.rs#L98

Overall, I'm not sure when to use `may_load` vs. `load`. Perhaps we can keep an eye on this and talk about it during a meeting as we learn. Looking great

mikedotexe (commented):
This looks right to me, but I think I'll wait a bit to dig in more on the difference between `may_load` and `load` and move to another PR for now

contracts/cw-croncat/src/agent.rs
mikedotexe: 👍🏼

PR #20, opened 2022-06-28
Tests for is_valid_msg

### LINKED ISSUE #16: [Tests] Task's `is_valid_msg`
This may be a good one to do after: https://github.com/CronCats/cw-croncat/issues/15.

In `types.rs` we see a function on the `Task` struct called `is_valid_msg`.

In Cosmos there different kinds of "Messages" (you can [read more here](https://docs.cosmwasm.com/docs/1.0/architecture/composition#messages)) but with Croncat we may not support all of them yet. For instance, as you can see there are several Message types that are not supported and therefore invalid.

https://github.com/CronCats/cw-croncat/blob/c63bd6268ba6e8e238c5439c30b3b749246e8256/packages/cw-croncat-core/src/types.rs#L187-L221

Let's write a test at the bottom of `types.rs` that ensures that valid Tasks return true, and also that invalid ones are caught. Let's have three different, valid Tasks, and five invalid tasks.

### Valid tests

- [ ] Task that's block-based, occurs once
- [ ] Task that's time-based, occurs once
- [ ] Task that's recurring

### Invalid tests

- [ ] A Task that is supposed to call the Croncat contract but is from Alice's account (The only tasks that are allowed to call Croncat must be sent from Croncat)
- [ ] A Task trying to use the `CosmosMsg::Bank` `Send` message type
- [ ] - [ ] A Task trying to use the `CosmosMsg::Bank` `Burn` message type
- [ ] - [ ] A Task trying to use the `CosmosMsg::Gov` `Vote` message type
- [ ] - [ ] A Task trying to use the `CosmosMsg::Ibc` `Transfer` message type

nino1415:
Closes #16

### DIFF
--- a/packages/cw-croncat-core/src/types.rs
+++ b/packages/cw-croncat-core/src/types.rs
@@ -449,3 +449,260 @@ impl Interval {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use cosmwasm_std::{IbcTimeout, VoteOption};
+
+    use super::*;
+
+    #[test]
+    fn is_valid_msg_once_block_based() {
+        let task = Task {
+            owner_id: Addr::unchecked("bob"),
+            interval: Interval::Once,
+            boundary: Boundary {
+                start: Some(BoundarySpec::Height(4)),
+                end: Some(BoundarySpec::Height(8)),
+            },
+            stop_on_fail: false,
+            total_deposit: Default::default(),
+            actions: vec![Action {
+                msg: CosmosMsg::Wasm(WasmMsg::Execute {
+                    contract_addr: "alice".to_string(),
+                    msg: Binary::from(vec![]),
+                    funds: vec![Coin::new(10, "coin")],
+                }),
+                gas_limit: Some(5),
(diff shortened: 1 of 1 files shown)


packages/cw-croncat-core/src/types.rs
```diff
@@ -453,8 +453,260 @@ impl Interval {
...
+        assert!(task.is_valid_msg(
+            &Addr::unchecked("alice2"),
+            &Addr::unchecked("bob"),
+            &Addr::unchecked("bob")
+        ));
+    }
+
+    #[test]
+    fn is_valid_msg_wrong_account() {
+        // A task with CosmosMsg::Gov Vote should return false
```
mikedotexe, line 555: I think this comment might be copy/pasted from another test)

mikedotexe, line 555:
```suggestion
        // Cannot create a task to execute on the cron manager when not the owner
```

mikedotexe (approved):
looking great. I just have the one thought on a test comment

PR #28, opened 2022-06-30
`update_settings` doesn't allow attached deposit

### LINKED ISSUE #23: `update_settings` method should not allow attached funds
Panic attached funds on update_settings method

In the `owner.rs` file we see the `update_settings` function:
https://github.com/CronCats/cw-croncat/blob/c63bd6268ba6e8e238c5439c30b3b749246e8256/contracts/cw-croncat/src/owner.rs#L42-L122

We want to ensure that there are no attached funds when someone calls this method.

Perhaps a good place to start understand how this works is to look at the `CreateTask` Execute Message, which is here, just like the rest of the Execute Messages:

https://github.com/CronCats/cw-croncat/blob/c63bd6268ba6e8e238c5439c30b3b749246e8256/contracts/cw-croncat/src/contract.rs#L116

This shows that when someone executes that Message, it'll actually call the function `create_task`, located here:
https://github.com/CronCats/cw-croncat/blob/c63bd6268ba6e8e238c5439c30b3b749246e8256/contracts/cw-croncat/src/tasks.rs#L219-L230

Notice in the above link there is a check for `info.funds`. This is likely the way you'll want to check for funds.

If anyone calls `update_settings` with funds it should panic with an error. If there doesn't exist the proper error, please add it to `error.rs`

nino1415:
Closes #23

`update_settings` doesn't allow attached funds (but allows attaching coins with zero amount).
Changed tests `update_settings`, `move_balances_auth_checks` and `move_balances_native` so that they don't attach deposit to `execute` with `ExecuteMsg::UpdateSettings`.

### DIFF
--- a/contracts/cw-croncat/src/owner.rs
+++ b/contracts/cw-croncat/src/owner.rs
@@ -290,6 +290,14 @@ mod tests {
             _ => panic!("Must return unauthorized error"),
         }
 
+        // non-zero deposit fails
+        let with_deposit_info = mock_info("owner_id", &coins(1000, "meow"));
+        let res_fail = store.execute(deps.as_mut(), mock_env(), with_deposit_info, payload.clone());
+        match res_fail {
+            Err(ContractError::AttachedDeposit {}) => {}
+            _ => panic!("Must return deposit error"),
+        }
+
         // do the right thing
         let res_exec = store
             .execute(deps.as_mut(), mock_env(), info.clone(), payload)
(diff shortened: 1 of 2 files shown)


contracts/cw-croncat/src/owner.rs
```diff
@@ -275,13 +283,29 @@ mod tests {
         };
 
         // non-owner fails
-        let unauth_info = mock_info("michael_scott", &coins(2, "shrute_bucks"));
+        let unauth_info = MessageInfo {
+            sender: Addr::unchecked("michael_scott"),
+            funds: vec![],
```
mikedotexe, line 288: makes sense, otherwise will return `Err(AttachedDeposit)` 👍🏼

mikedotexe (approved):

PR #30, opened 2022-07-01
`query_get_tasks` reimplemented using iterator pattern and tests that pagination works as expected

### LINKED ISSUE #24: [Tests] `query_get_tasks`: Confirm pagination works as expected
We have this function in `tasks.rs`:
https://github.com/CronCats/cw-croncat/blob/c63bd6268ba6e8e238c5439c30b3b749246e8256/contracts/cw-croncat/src/tasks.rs#L14-L59

Note that it takes `from_index` and `limit` which basically means, "start at this index and give me X"

Note that we should always use Rust's iterator pattern, which I believe we're all familiar with. If you want to hear more about it, take a listen for a couple minutes starting at the YouTube video linked here:
https://twitter.com/mikedotexe/status/1538378440436355073

Note that we have some tests in this `tasks.rs` file but none of them use the `from_index` or `limit` parameters.
For instance, this one doesn't use it:
https://github.com/CronCats/cw-croncat/blob/c63bd6268ba6e8e238c5439c30b3b749246e8256/contracts/cw-croncat/src/tasks.rs#L660-L671

So this ticket will be writing a test that ensures people can give `None`/`None`, or `Some(6)`/`None` to start at index 6 and get all the rest, or `Some(6)`/`Some(10)` to start at index 6 and get the next 10. We'll also want to make sure if there are only 9 tasks and someone says to "take 10" that it doesn't panic, but will return as many as are possible.

## <a href="#" id="manual-anchor-bro">`cw-plus` enumeration reference</a>

This is a good reference to pagination in `cw-plus`:
```rs
pub fn query_all_accounts(
    deps: Deps,
    start_after: Option<String>,
    limit: Option<u32>,
) -> StdResult<AllAccountsResponse> {
    let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
    let start = start_after.map(|s| Bound::ExclusiveRaw(s.into()));

    let accounts = BALANCES
        .keys(deps.storage, start, None, Order::Ascending)
        .take(limit)
        .map(|item| item.map(Into::into))
        .collect::<StdResult<_>>()?;

    Ok(AllAccountsResponse { accounts })
}
```
https://github.com/CosmWasm/cw-plus/blob/9d0f06d282bdc884252bb2dbb060c926438df0ed/contracts/cw20-base/src/enumerable.rs#L36-L51

I think the idea of `start_after` might be a good idea, and perhaps better than `from_index`
Let's discuss in Element further, too

Buckram123:
Closes #24.

Played a bit with using `from_hash`, instead of `from_index`. And good news - it seems like elements sorted by "key". Example, why it's good: 
Let's say we have 5 hashes: ["a", "b", "c", "d", "e"] and we iterate with `limit = 2`, last elements we got was ["a", "b"]. And imagine after that "b" gets removed:

using `from_index(2)` we get: ["d", "e"],
using `from_hash("b")` we get: ["c", "d"].

It's easy to change to `from_hash`, so just let me know if we want that behavior instead

### DIFF
--- a/contracts/cw-croncat/src/tasks.rs
+++ b/contracts/cw-croncat/src/tasks.rs
@@ -22,27 +22,16 @@ impl<'a> CwCroncat<'a> {
         let size: u64 = self
             .task_total
             .may_load(deps.storage)?
-            .unwrap_or(100)
+            .unwrap_or(0)
             .min(1000);
-        if let Some(index) = from_index {
-            start = index;
-        }
-        let mut end = u64::min(start.saturating_add(100), size);
-        if let Some(l) = limit {
-            end = u64::min(start.saturating_add(l), size);
-        }
-
-        // NOTE: could setup another index to allow efficient paginated
-        let keys: Vec<Vec<u8>> = self
-            .tasks
-            .keys(deps.storage, None, None, Order::Ascending)
-            .collect::<StdResult<Vec<_>>>()?;
-
-        for i in start..end {
-            let task_hash = &keys[i as usize];
-            let res = self.tasks.may_load(deps.storage, task_hash.to_vec())?;
-            if let Some(task) = res {
-                ret.push(TaskResponse {
+        let from_index = from_index.unwrap_or_default();
+        let limit = limit.unwrap_or(100).min(size);
+        self.tasks
+            .range(deps.storage, None, None, Order::Ascending)
+            .skip(from_index as usize)
+            .take(limit as usize)
+            .map(|res| {
+                res.map(|(_k, task)| TaskResponse {
                     task_hash: task.to_hash(),
                     owner_id: task.owner_id,
                     interval: task.interval,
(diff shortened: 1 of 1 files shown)


contracts/cw-croncat/src/tasks.rs
```diff
@@ -17,32 +17,19 @@ impl<'a> CwCroncat<'a> {
...
-            let task_hash = &keys[i as usize];
-            let res = self.tasks.may_load(deps.storage, task_hash.to_vec())?;
-            if let Some(task) = res {
-                ret.push(TaskResponse {
+        let from_index = from_index.unwrap_or_default();
+        let limit = limit.unwrap_or(100).min(size);
+        self.tasks
+            .range(deps.storage, None, None, Order::Ascending)
+            .skip(from_index as usize)
+            .take(limit as usize)
```
mikedotexe, line 30: 🔥

mikedotexe (approved):

mikedotexe:
This is great, and I think it addresses what the ticket was talking about regarding the `start_after` approach.

It's going to be interesting to see if we can have this consistent pattern in our code, and I'm hoping we can avoid confusion to any user-facing functions that may also have the idea of "start after X and get this many" in all the places it may apply.

PR #31, opened 2022-07-01
Replacing tuples/tuple structs in Query Responses

Buckram123:
`serialize_tuple_struct()` not implemented in `serde-json-wasm`, so it will fail.
And tuples hard to read if there is multiple values.


packages/cw-croncat-core/src/msg.rs
```diff
@@ -147,7 +147,19 @@ pub struct GetAgentIdsResponse {
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
-pub struct GetAgentTasksResponse(pub u64, pub u64);
+pub struct GetAgentTasksResponse {
+    pub availible_tasks: u64,
+    pub next_slot: u64,
```
Buckram123, line 152: Not sure if I understood right those values, what `get_agent_tasks()` query should return in the end @mikedotexe ?

Issue #26, opened 2022-06-29
Implement `send_base_agent_reward`


mikedotexe:
This needs implementation - commented code shows whats needed. The comments are from the NEAR Protocol code, so it will be different. (For instance we won't use `U128` etc.)

https://github.com/CronCats/cw-croncat/blob/c63bd6268ba6e8e238c5439c30b3b749246e8256/contracts/cw-croncat/src/manager.rs#L290-L307

I want to note that the comment here says:

>/// Used in cases where there are empty slots or failed txns

I believe this should only apply for failed transactions, since we will be telling agents about the upcoming tasks. I believe there will be many empty slots and that no longer applies for this, improved CosmWasm implementation of Croncat.

PR #33, opened 2022-07-08
Sending agent rewards

deveusss:
Closes [#26](https://github.com/CronCats/cw-croncat/issues/26)
Internal management of agent reward
Used in cases where there are empty slots or failed txns
Keep the agent profitable, as this will be a business expense

### DIFF
--- a/contracts/cw-croncat/src/manager.rs
+++ b/contracts/cw-croncat/src/manager.rs
@@ -1,8 +1,10 @@
 use crate::error::ContractError;
 use crate::state::{Config, CwCroncat, QueueItem};
+use cosmwasm_std::Coin;
 use cosmwasm_std::{
     Addr, DepsMut, Empty, Env, MessageInfo, Reply, Response, StdResult, Storage, SubMsg,
 };
+use cw20::Balance;
 use cw_croncat_core::types::{Agent, SlotType};
 
 impl<'a> CwCroncat<'a> {
@@ -48,7 +50,7 @@ impl<'a> CwCroncat<'a> {
         // if empty slot found, let agent get paid for helping keep house clean
         let slot = self.get_current_slot_items(&env.block, deps.storage);
         if slot.is_none() {
-            self.send_base_agent_reward(deps.storage, agent);
+            self.send_base_agent_reward(deps.storage, agent,info);
             return Err(ContractError::CustomError {
                 val: "No Tasks For Slot".to_string(),
             });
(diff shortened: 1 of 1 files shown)


contracts/cw-croncat/src/manager.rs
```diff
@@ -290,19 +291,40 @@ impl<'a> CwCroncat<'a> {
...
+        mut _agent: Agent,
+        message: MessageInfo,
+    ) {
+        let mut config: Config = self.config.load(_storage).unwrap();
+
+        let agent_base_fee = config.agent_fee.clone();
+        let coin = vec![agent_base_fee.clone()];
+        let add_native: Balance = Balance::from(coin);
+
+        _agent.balance.add_tokens(add_native.clone());
```
mikedotexe, line 306: I think we'll want to have `_storage` and `_agent` be `storage` and `agent`, removing the underscores.

>Note: Identifiers starting with an underscore are typically used to indicate an identifier that is intentionally unused, and will silence the unused warning in rustc.

https://doc.rust-lang.org/reference/identifiers.html?highlight=underscore#identifiers

Like if I add inside this function:

```rs
let mike = "hey friend";
```

and try to run

    ./build.sh

I get:

Screen Shot 2022-07-11 at 5 51 35 PM

So I believe we only want to prefix with underscore (`_`) when we're not going to use the variable, but here we do use the variable.

contracts/cw-croncat/src/manager.rs
```diff
@@ -290,19 +291,40 @@ impl<'a> CwCroncat<'a> {
...
+        _agent.total_tasks_executed = _agent.total_tasks_executed.saturating_add(1);
+        println!("{:?}", add_native);
+        println!("{:?}", config.available_balance.native);
+
+        if !config.available_balance.native.is_empty()
+            && config.available_balance.native.first().unwrap().amount >= agent_base_fee.amount
+        {
+            config.available_balance.minus_tokens(add_native);
+        } else {
+            panic!("Not enough available balance for sending agent reward");
```
mikedotexe, line 316: I think we'll want to add this error and the checks on `Config`'s available balance in the `proxy_call` function in `manager.rs`. Looks like that is the only function that calls this private function (`send_base_agent_reward`) and it has a couple similar checks, like this one:

https://github.com/CronCats/cw-croncat/blob/f4683ad46ef5660be691fddb844fed4c8ce6c53f/contracts/cw-croncat/src/manager.rs#L24-L29

I believe we'll always try to use the pattern of `Err(ContractError::SOMETHING…` instead of using the `panic!` macro when throwing errors.

mikedotexe (changes requested):
Couple of small things mentioned in the comments)) Thanks!

PR #34, opened 2022-07-11
A script for deploying on testnet

nino1415:
The script deploys on uni-3 testnet, registers an agent, creates a task and executes it. 
OWNER, AGENT, USER should be replaced with your wallets.

### DIFF
--- a/deploy_testnet_example.sh
+++ b/deploy_testnet_example.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+set -e
+
+cargo wasm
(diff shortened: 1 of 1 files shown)


deploy_testnet_example.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+set -e
+
+cargo wasm
+# In case of M1 MacBook use rust-optimizer-arm64 instead of rust-optimizer
```
mikedotexe, line 5: Niiice

mikedotexe (approved):

PR #36, opened 2022-07-13
Fix pop_slot_item

nino1415:
For now, if we create several tasks and call `proxy_call`, it succeeds for the first time and fails if we try to call it again. 

It happens because these tasks are stored under separate keys in `block_slots: time_slots: Map<'a, u64, Vec<Vec<u8>>>` (or `time_slots`), and calling `pop_slot_item` should delete one of the keys, so that second `proxy_call` could take the remaining key and hash and execute the second task. Instead `pop_slot_item` leaves the first key with empty vector and the second `proxy_call` fails to get the correct `slot_id`.

That is why I think we should delete a key from `block_slots` (`time_slots`) if the corresponding vector of hashes is empty.

Also added a test, where three tasks are created and `proxy_call` is called on each of them.

### DIFF
--- a/contracts/cw-croncat/src/manager.rs
+++ b/contracts/cw-croncat/src/manager.rs
@@ -1199,4 +1199,149 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn c() -> StdResult<()> {
+        let (mut app, cw_template_contract) = proper_instantiate();
+        let contract_addr = cw_template_contract.addr();
+        let proxy_call_msg = ExecuteMsg::ProxyCall {};
+
+        // Doing this msg since its the easiest to guarantee success in reply
+        let msg = CosmosMsg::Wasm(WasmMsg::Execute {
+            contract_addr: contract_addr.to_string(),
+            msg: to_binary(&ExecuteMsg::WithdrawReward {})?,
+            funds: coins(1, NATIVE_DENOM),
+        });
+
+        let msg2 = CosmosMsg::Wasm(WasmMsg::Execute {
+            contract_addr: contract_addr.to_string(),
+            msg: to_binary(&ExecuteMsg::WithdrawReward {})?,
+            funds: coins(2, NATIVE_DENOM),
+        });
+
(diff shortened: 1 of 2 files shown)


contracts/cw-croncat/src/slots.rs
```diff
@@ -214,16 +214,13 @@ impl<'a> CwCroncat<'a> {
         // Need to remove this slot if no hash's left
         if slot_data.is_empty() {
             self.clean_slot(storage, slot, kind);
-        }
-
-        if hash.is_some() {
+        } else {
             store
                 .update(storage, *slot, |_d| -> StdResult<Vec<_>> { Ok(slot_data) })
                 .ok();
```
Buckram123, line 220: Since we touching this method, I think it's cleaner to rewrite as 
```suggestion
            store.save(storage, *slot, &slot_data).ok()?;
```

contracts/cw-croncat/src/slots.rs
```diff
@@ -214,16 +214,11 @@ impl<'a> CwCroncat<'a> {
...
         }
 
-        if hash.is_some() {
-            store
-                .update(storage, *slot, |_d| -> StdResult<Vec<_>> { Ok(slot_data) })
-                .ok();
-            return hash;
-        }
-
-        None
```
Buckram123, line 226: Huh, so when we had last hash pop from this slot we override "clean_slot", so it's active again with 0 elements. Took some time for me to understand it :D

contracts/cw-croncat/src/manager.rs
```diff
@@ -1199,4 +1199,149 @@ mod tests {
...
+        app.execute_contract(
+            Addr::unchecked(ADMIN),
+            contract_addr.clone(),
+            &create_task_msg2,
+            &coins(10, NATIVE_DENOM),
+        )
+        .unwrap();
+
+        // the third task is created in another block
+        app.update_block(add_little_time);
```
Buckram123, line 1294: Nice

Buckram123 (approved):

PR #40, opened 2022-07-22
Scripts for deploying and instantiating to mainnet

nino1415:
Added scripts for deploying and instantiating to juno-1.

To run `deploy_mainnet.sh` one would need to provide wallet/address of the owner.
To run `instantiate_mainnet.sh` one would need to provide wallet/address and `CODE_ID`.
For example:
```
sh deploy_mainnet.sh owner
sh instantiate_mainnet.sh owner 556
```

### DIFF
--- a/deploy_mainnet.sh
+++ b/deploy_mainnet.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -e
+
+sh build.sh
+# In case of M1 MacBook use rust-optimizer-arm64 instead of rust-optimizer
+docker run --rm -v "$(pwd)":/code \
+  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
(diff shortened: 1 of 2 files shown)


deploy_mainnet.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -e
+
+sh build.sh
+# In case of M1 MacBook use rust-optimizer-arm64 instead of rust-optimizer
+docker run --rm -v "$(pwd)":/code \
+  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+  cosmwasm/rust-optimizer:0.12.6
```
Buckram123, line 9: Saw DA0-DA0 adding `platform` flag here, wonder if it's for M1 https://github.com/DA0-DA0/dao-contracts/blob/main/scripts/deploy_local.sh#L42

PR #44, opened 2022-07-26
Refactor Boundary and added validation for it

### LINKED ISSUE #43: Refactor Boundary and add validation
Currently we have this struct layout:
https://github.com/CronCats/cw-croncat/blob/a869233015e654b45523d4dfa4bfa35b1c60d56e/packages/cw-croncat-core/src/types.rs#L85-L100

And it has two main issues:
1. `start` and `end` can be different `BoundarySpec` enum variant(for ex. `start` - `Time` and `end` - `Height`).
2.  Since we know that `BoundarySpec` is bound to `Interval` variants, we also waste some compute time(and complicate readability) pattern matching it in these methods:
https://github.com/CronCats/cw-croncat/blob/a869233015e654b45523d4dfa4bfa35b1c60d56e/packages/cw-croncat-core/src/types.rs#L355
https://github.com/CronCats/cw-croncat/blob/a869233015e654b45523d4dfa4bfa35b1c60d56e/packages/cw-croncat-core/src/types.rs#L311

and here: 
https://github.com/CronCats/cw-croncat/blob/a869233015e654b45523d4dfa4bfa35b1c60d56e/packages/cw-croncat-core/src/types.rs#L419-L433

So my idea is to replace it with 
```rust

pub enum Boundary {
    Height {
        start: Option<Uint64>,
        end: Option<Uint64>,
    },
    Time {
        start: Option<Timestamp>,
        end: Option<Timestamp>,
    },
}

pub struct BoundaryValidated {
    start: Option<u64>,
    end: Option<u64>,
}
```
The first one we use in `TaskRequest` and `TaskResponse`, and the second  we store in our `Task` struct

Buckram123:
Closes #43

### DIFF
--- a/contracts/cw-croncat/src/slots.rs
+++ b/contracts/cw-croncat/src/slots.rs
@@ -1,159 +1,7 @@
-use crate::traits::IntervalExt;
 use crate::CwCroncat;
-use cosmwasm_std::{BlockInfo, Env, Order, StdResult, Storage};
-use cron_schedule::Schedule;
+use cosmwasm_std::{BlockInfo, Order, StdResult, Storage};
 pub use cw_croncat_core::types::Interval;
-use cw_croncat_core::types::{Boundary, BoundarySpec, SlotType};
-use std::str::FromStr;
-
-fn get_next_block_limited(env: Env, boundary: Boundary) -> (u64, SlotType) {
-    let current_block_height = env.block.height;
-
-    let next_block_height = if boundary.start.is_some() {
-        match boundary.start.unwrap() {
-            // Note: Not bothering with time, as that should get handled with the cron situations,
-            // and probably throw an error when mixing blocks and cron
-            BoundarySpec::Height(id) => {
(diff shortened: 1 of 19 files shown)


contracts/cw-croncat/src/slots.rs
```diff
@@ -1,159 +1,7 @@
...
-            Interval::Once => true,
-            Interval::Immediate => true,
-            Interval::Block(_) => true,
-            Interval::Cron(crontab) => {
-                let s = Schedule::from_str(crontab);
-                s.is_ok()
-            }
-        }
-    }
-}
```
Buckram123, line 156: This was not used, (it got moved to `packages/cw-croncat-core/src/types.rs`)

Buckram123: > This was not used, (it got moved to `packages/cw-croncat-core/src/types.rs`)

Also there is main changes to `Boundary`

PR #46, opened 2022-07-26
Modify is_valid_msg so preliminary Bank msg werks

mikedotexe:
We need to make sure users can send a `Bank` message, since this is the primary case we're trying to solve for DAO DAO.

I've done some very basic work here to do some sanity checks on what the user (the one creating the task) is trying to accomplish, and make sure they aren't doing any funny business.

Note that this will only work for Croncat tasks that are "immediate" instead of recurring ones, because it does not implement the check to see if the user attached enough for multiple payments.

Also, obviously note that this is not production code since we're hardcoding in the JUNO testnet token denomination, which is `ujunox`. Obviously that part will not make it into production, but I really want us to see a Bank message working on testnet. It SHOULD WORK, but if it doens't work, we should stop what we're doing and focus on why a simple Bank message isn't working, and prioritize that. If it DOES WORK then we can continue making this production ready, remove the hardcoding, allow for recurring tasks, etc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "0.7.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.57"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -214,6 +223,7 @@ dependencies = [
  "cw20",
  "cw20-base",
  "hex",
+ "regex",
  "schemars",
  "serde",
  "sha2",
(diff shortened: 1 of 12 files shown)


mikedotexe:
We can try this out on testnet by using `junod`. Some helpful stuff here:
https://docs.junonetwork.io/cli/useful-cli-commands

When I run:

    junod keys list

I get:

```sh
- name: deployer
  type: local
  address: juno1kr3q60whahcqr785x405y4qkc0lm8ywwq5u9c6
  pubkey: '{"@type":"/cosmos.crypto.secp256k1.PubKey","key":"A4M3B3fpt5iieM0KlrP5HiRtyQuLSw5g81xDZMtygixC"}'
  mnemonic: ""
- name: mike
  type: local
  address: juno1hemfted3mgj04hae507lwhm8xuj33hqjg6y9q8
  pubkey: '{"@type":"/cosmos.crypto.secp256k1.PubKey","key":"A93L1fxTjcvwOAqTPY5USq8nPPAge83YCJo2eVxriONC"}'
  mnemonic: ""
```

Then I can go into the [Juno faucet Discord](https://discord.com/channels/816256689078403103/842073995059003422) if I need some tokens and type:

    $request juno1hemfted3mgj04hae507lwhm8xuj33hqjg6y9q8

Then I think we should be able to use this command, of course you'll want to change a couple parameters:

    junod tx wasm execute juno14lhj9w77l968a2zpf9hqdscrsydvsp3e9sx5ekzpe2a658e2s9tq0eyx0v "$SEND" --amount 1000000ujunox --from mike --node https://rpc.uni.juno.deuslabs.fi:443 --chain-id uni-3 --gas-prices 0.025ujunox --gas-adjustment 1.3 --broadcast-mode block

PR #51, opened 2022-08-02
Support of Cw20 for tasks

### LINKED ISSUE #45: Implement cw20 receiver
To support tasks, that is related to cw20 we have to implement this:
https://github.com/CosmWasm/cw-plus/blob/main/packages/cw20/README.md#receiver

and everything that comes from it.

Buckram123:
To fully support cw20 we need:

- [x] Implement receiver. 
Current implementation:
- we have a list of user cw20 `balances`
- user can add tokens with `Cw20ExecuteMsg::Send`
- `balances` gets subtracted when user creates task with non-empty `total_cw20_deposit`, or refills task with it.
Closes #45 
- [x] Refund back to user cw20 `balances`, when task gets removed.
- [x] Validate task has enough `total_cw20_deposit` for every `action`.
- [x] ~~Handle callback from it.~~ (Gonna do it in a different PR, after we fully decide how we handle transferred coins #55)
- [x] Test coverage


Buckram123:
Some unmentioned features of cw20 yet:
- [x] Added withdraw method
- [x] Added query method for checking user cw20 balances inside this contract

PR #52, opened 2022-08-02
Refactor generic balance

Buckram123:
I needed `checked_sub` method for `GenericBalance` in #51, so I decided to refactor it as well.

Improvements
1. In CosmWasm it's not idiomatic to panic, so I replaced `sub`/`add` methods with `checked_add`/`checked_sub`, that returns `Result`s
2. Replaced `add_tokens` with `checked_add_native ` and `checked_add_cw20`, same for `sub_tokens`. 

- Reason: redundant pattern match. Don't think we will ever have a situation where we don't know what type are we adding/subbing. And even if we don't know the type we adding/subbing imo it's better to match for this unique case

3.  "rhs" taking references now, so we have less clones


NinoLipartiia (commented):
Thanks Misha, seems like a useful refactor!
I think we should we delete redundant `EmptyBalance` from `ContractError`, since it is already in `CoreError`
https://github.com/CronCats/cw-croncat/blob/85d60d6d32964fb9e24de4bd51f86e815eab12b4/contracts/cw-croncat/src/error.rs#L18-L19

NinoLipartiia (approved):

PR #59, opened 2022-08-10
Supporting multiaction for tasks and balance fixes

### LINKED ISSUE #54: Support multiple actions
Currently if we try to create a task with multiple actions it will fail in `reply()` here: 
https://github.com/CronCats/cw-croncat/blob/a00b5780b9f4424577554d39a54fbed1065a546b/contracts/cw-croncat/src/contract.rs#L184-L185
Notice here `next_idx` stays the same, so after first reply second will fail this transaction
https://github.com/CronCats/cw-croncat/blob/a00b5780b9f4424577554d39a54fbed1065a546b/contracts/cw-croncat/src/manager.rs#L186-L193

Not sure yet what we want to do here, but few ideas:

1. Remove `QueueItem` only after it's the last response? (in short, we add a counter corresponding to the number of actions)
2. increment `next_idx` in a `proxy_call()`, so it's different `QueueItem`'s. There is actually currently unused field `prev_idx`

Buckram123:
Closes #54.
New stuff:

- Support of multiple action in one task. 
  - Only the last action will call `proxy_callback()`, and actions before that used to check used deposit in cases of transfers
- Validating if task has enough deposit to reschedule this task.
- Calculating amount of used tokens at `create_task()`, merged it with `is_valid_msg()`, so we don't iterate `actions` multiple times
  - Storing it in the task itself. What you think, should we show `used_deposit` in the `TaskResponse`?
  - Thinking about better naming for `is_valid_msg()` now? What you think about `is_valid_msg_calculate_usage()`
- Calculating gas usage and agent_base_reward right in the `proxy_call`
- Calculate task balance changes in the reply(so we can properly change it after TX is passed)

Fixes:

- After agent withdraw his balances we decrease his balance in the state(CRITICAL, infinite withdraw)
- Wrong calculation of agent base reward
- `send_base_agent_reward()` didn't sent any funds, cause it was followed by Err(), which rollbacks the state.
- Refunds if task balance is empty at `remove_task()`. If we try to send zero coins it will fail the tx and rollback state(and we really don't want this happen during the `proxy_call()`) 
- Removed  redundant clones
- Some minor refactors

### DIFF
--- a/contracts/cw-croncat/src/state.rs
+++ b/contracts/cw-croncat/src/state.rs
@@ -1,4 +1,4 @@
-use crate::balancer::RoundRobinBalancer;
+use crate::{balancer::RoundRobinBalancer, ContractError};
 use cosmwasm_std::{Addr, Coin, StdResult, Storage, Timestamp};
 use cw20::Cw20CoinVerified;
 use cw_storage_plus::{Index, IndexList, IndexedMap, Item, Map, MultiIndex};
@@ -52,10 +52,15 @@ pub struct QueueItem {
     // This is used to track disjointed callbacks
     // could help scheduling multiple calls across txns
     // could help for IBC non-block bound txns
-    pub prev_idx: Option<u64>,
+    // not used yet, need more discover
+    // pub prev_idx: Option<u64>,
+
+    // counter of actions helps track what type of action it is
+    pub action_idx: u64,
     pub task_hash: Option<Vec<u8>>,
     pub task_is_extra: Option<bool>,
     pub agent_id: Option<Addr>,
+    pub failed: bool,
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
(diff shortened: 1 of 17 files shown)


contracts/cw-croncat/src/state.rs
```diff
@@ -176,6 +181,24 @@ impl<'a> CwCroncat<'a> {
...
+        failed: bool,
+    ) -> Result<QueueItem, ContractError> {
+        self.reply_queue.update(storage, idx, |rq| {
+            let mut rq = rq.ok_or(ContractError::UnknownReplyID {})?;
+            // if first fails it means whole thing failed
+            // for cases where we stop task on failure
+            if !rq.failed {
+                rq.failed = failed;
+            }
+            rq.action_idx += 1;
```
Buckram123, line 198: we have this counter, that goes up from every action, so we can check what type of action it is and do `proxy_callback()` at  the last action

contracts/cw-croncat/src/tasks.rs
```diff
@@ -410,21 +408,24 @@ impl<'a> CwCroncat<'a> {
...
-            .add_submessage(submsgs))
+        self.config
+            .update(storage, |mut c| -> Result<_, ContractError> {
+                c.available_balance
+                    .checked_sub_native(&task.total_deposit)?;
+                Ok(c)
+            })?;
+        // setup sub-msgs for returning any remaining total_deposit to the owner
+        if !task.total_deposit.is_empty() {
+            Ok(Response::new()
```
Buckram123, line 420: if we transfer zero it will fail and we don't want the whole `proxy_call()` to fail cause we sent zero

Buckram123 (commented):
Quick overview of highlights

deveusss (approved):

PR #71, opened 2022-08-19
Proxy call updates

NinoLipartiia:
Added `tasks_with_rules`, `tasks_with_rules_total`, `block_slot_rules` and `time_slot_rules` to `CwCroncat`
`proxy_call`  takes optional parameter `task_hash`
Also there is a new query `get_tasks_with_rules`


contracts/cw-croncat/src/contract.rs
```diff
@@ -143,7 +143,10 @@ impl<'a> CwCroncat<'a> {
                 task_hash,
                 cw20_coins,
             } => self.refill_task_cw20(deps, info, task_hash, cw20_coins),
-            ExecuteMsg::ProxyCall {} => self.proxy_call(deps, info, env),
+            ExecuteMsg::ProxyCall {
```
mikedotexe, line 146: Whoa, I didn't realize you could use `match` like this where you can tell whether the param is `Some` or `None`

contracts/cw-croncat/src/manager.rs
```diff
@@ -211,6 +177,105 @@ impl<'a> CwCroncat<'a> {
...
+            return Err(ContractError::NoTaskFound {});
+        }
+        let task = some_task.unwrap();
+
+        self.check_bank_msg(deps.as_ref(), &info, &env, &task)?;
+
+        // Check rules
+        let rules = task.rules.as_ref().expect("No rules");
+        for rule in rules {
+            let res: (bool, Option<String>) =
```
mikedotexe, line 242: awesome.

contracts/cw-croncat/src/manager.rs
```diff
@@ -211,6 +177,105 @@ impl<'a> CwCroncat<'a> {
...
+                deps.querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+                    contract_addr: rule.contract_addr.clone().to_string(),
+                    msg: rule.msg.clone(),
+                }))?;
+            if !res.0 {
+                return Err(ContractError::RulesNotReady {});
```
mikedotexe, line 248: Might be nice to tell which rule failed. Like passing an index into `RulesNotReady` so the error can say…

```rs
#[error("Rules are not ready. Failed at rule {index:?}.")]
RulesNotReady { index: u8 },
```

contracts/cw-croncat/src/manager.rs
```diff
@@ -211,6 +177,105 @@ impl<'a> CwCroncat<'a> {
...
+            env.block.time.nanos() >= time
+        } else {
+            // This shouldn't happen
+            return Err(ContractError::CustomError {
+                val: "Task doesn't have block or time slot".to_string(),
+            });
+        };
+        if !task_ready {
+            return Err(ContractError::CustomError {
+                val: "Wrong slot for this task".to_string(),
```
mikedotexe, line 271: Feels like we could have a different error message, unless I'm misunderstanding:

```suggestion
                val: "Task is not ready".to_string(),
```

contracts/cw-croncat/src/tasks.rs
```diff
@@ -32,6 +34,27 @@ impl<'a> CwCroncat<'a> {
...
+    /// Used by the frontend for viewing tasks
+    pub(crate) fn query_get_tasks_with_rules(
+        &self,
+        deps: Deps,
+        from_index: Option<u64>,
+        limit: Option<u64>,
+    ) -> StdResult<Vec<TaskWithRulesResponse>> {
+        let size: u64 = self.tasks_with_rules_total.load(deps.storage)?.min(1000);
+        let from_index = from_index.unwrap_or_default();
+        let limit = limit.unwrap_or(100).min(size);
```
mikedotexe, line 49: We should probably make `100` a value we can change in `Config`, but not a big deal at this point.

mikedotexe (approved):

contracts/cw-croncat/src/manager.rs
NinoLipartiia: We have `Err(ContractError::RulesNotReady {})` if one of the rules failed. But here we were checking whether the task can be executed in the current clot, so I think the error should be `Wrong slot for this task` or something similar

PR #77, opened 2022-08-23
Fix balance methods in cw-rules

NinoLipartiia:
Changed `query_has_balance` and `query_get_cw20_balance` so that they return `Coin` in `Binary`.
Renamed `query_has_balance` to `query_has_balance_gt` and implemented checking that the given address has more coins than the given balance.

Also I'm not sure whether `query_get_cw20_balance` should return `Coin`.  I think `Cw20CoinVerified` is more suitable as it has `address` instead of `denom`:
```
pub struct Cw20CoinVerified {
    pub address: Addr,
    pub amount: Uint128,
}
```
Another option is to return `Balance` in both `query_has_balance` and `query_get_cw20_balance`
```
pub enum Balance {
    Native(NativeBalance),
    Cw20(Cw20CoinVerified),
}
```

### DIFF
--- a/contracts/cw-rules/src/contract.rs
+++ b/contracts/cw-rules/src/contract.rs
@@ -127,16 +127,9 @@ fn query_has_balance_gt(
         Balance::Native(required_native) => {
             let balances = deps.querier.query_all_balances(valid_address)?;
             let required_vec = required_native.into_vec();
-            let mut has_balance = true;
-            for required_coin in required_vec {
-                if (required_coin.amount != Uint128::zero())
-                    && (!has_coins(&balances, &required_coin))
-                {
-                    has_balance = false;
-                    break;
-                }
-            }
-            has_balance
+            required_vec.iter().all(|required| {
+                required.amount == Uint128::zero() || has_coins(&balances, required)
+            })
         }
         Balance::Cw20(required_cw20) => {
             let balance_response: BalanceResponse = deps.querier.query_wasm_smart(
--- a/contracts/cw-rules/src/contract.rs
+++ b/contracts/cw-rules/src/contract.rs
@@ -5,7 +5,8 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    has_coins, to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response, StdResult, Uint128,
+    coin, has_coins, to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response, StdResult,
+    Uint128,
 };
 use cw2::set_contract_version;
 use cw20::{Balance, BalanceResponse};
@@ -100,15 +101,11 @@ fn query_get_cw20_balance(
     address: String,
 ) -> StdResult<RuleResponse<Option<Binary>>> {
     let valid_cw20 = deps.api.addr_validate(&cw20_contract)?;
-    let balance: BalanceResponse = deps
+    let balance_response: BalanceResponse = deps
         .querier
         .query_wasm_smart(valid_cw20, &cw20::Cw20QueryMsg::Balance { address })?;
-    let amount = if balance.balance.is_zero() {
-        None
-    } else {
-        Some(to_binary(&balance.balance)?)
-    };
-    Ok((true, amount))
+    let coin = coin(balance_response.balance.into(), cw20_contract);
+    Ok((true, to_binary(&coin).ok()))
 }
 
 fn query_has_balance_gt(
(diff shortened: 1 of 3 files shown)


contracts/cw-rules/src/msg.rs
```diff
@@ -27,8 +27,8 @@ pub enum QueryMsg {
         cw20_contract: String,
         address: String,
     },
-    HasBalance {
-        balance: Balance,
+    HasBalanceGT {
```
Buckram123, line 30: I think we want here `HasBalanceGt`, because snake_cased version of it will be `has_balance_g_t`, instead of `has_balance_gt`

contracts/cw-rules/src/contract.rs
```diff
@@ -114,30 +110,41 @@ fn query_get_cw20_balance(
...
+            let mut has_balance = true;
+            for required_coin in required_vec {
+                if (required_coin.amount != Uint128::zero())
+                    && (!has_coins(&balances, &required_coin))
+                {
+                    has_balance = false;
+                    break;
+                }
+            }
+            has_balance
```
Buckram123, line 139:
```suggestion
            required_vec.iter().all(|required| has_coins(&balances, required))
```

Buckram123 (changes requested):
Hey, good job on fixing query balances methods. Agree that `query_get_cw20_balance` should return `Cw20CoinVerified`

contracts/cw-rules/src/contract.rs
NinoLipartiia: Changed it to
```
required_vec.iter().all(|required| {
    required.amount == Uint128::zero() || has_coins(&balances, required)
})
```
to cover a case when required balance is zero

Buckram123 (approved):

contracts/cw-rules/src/contract.rs
```diff
@@ -114,30 +110,34 @@ fn query_get_cw20_balance(
...
-        Some(to_binary(&balance.balance)?)
-    };
-    Ok((true, amount))
+    let coin = coin(balance_response.balance.into(), cw20_contract);
+    Ok((true, to_binary(&coin).ok()))
 }
 
-fn query_has_balance(
-    balance: Balance,
+fn query_has_balance_gt(
```
mikedotexe, line 120: 👍🏼

contracts/cw-rules/src/contract.rs
```diff
@@ -114,30 +110,34 @@ fn query_get_cw20_balance(
...
-    let res = match (balance, required_balance) {
-        (Balance::Native(current), Balance::Native(expected)) => {
-            expected.0.iter().all(|c| has_coins(&current.0, c))
+    let valid_address = deps.api.addr_validate(&address)?;
+    let res = match required_balance {
+        Balance::Native(required_native) => {
+            let balances = deps.querier.query_all_balances(valid_address)?;
+            let required_vec = required_native.into_vec();
+            required_vec.iter().all(|required| {
+                required.amount == Uint128::zero() || has_coins(&balances, required)
```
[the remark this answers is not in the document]
mikedotexe: I also think we may want to make a small change so the function name is `query_has_balance_gte` (for greater-than-or-equal) since we're using `>=` here, and so does `has_coins`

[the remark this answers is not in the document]
NinoLipartiia: If you query "you must have 19 ujuno", then `required.amount` is 19, so it won't return true.

`required.amount == Uint128::zero()` is here for the case when you query "you must have at least 0 ujuno", then it should return true, even if address doesn't have any ujuno.

contracts/cw-rules/src/contract.rs
```diff
@@ -114,30 +110,34 @@ fn query_get_cw20_balance(
     address: String,
 ) -> StdResult<RuleResponse<Option<Binary>>> {
     let valid_cw20 = deps.api.addr_validate(&cw20_contract)?;
-    let balance: BalanceResponse = deps
+    let balance_response: BalanceResponse = deps
```
mikedotexe, line 113: Actually, can we also `addr_validate` the `address`? I think it'll be best to validate both the `cw20_contract` and that, otherwise people may get odd errors.

mikedotexe (changes requested):

contracts/cw-rules/src/contract.rs
mikedotexe: Other than that, this is looking amazing, and wonderful job with the tests as usual

mikedotexe: actually, let's just have this be a covered in this smaller ticket https://github.com/CronCats/cw-croncat/issues/86

mikedotexe (approved):

PR #103, opened 2022-09-16
feat: Improving scripts for local and testnet

deveusss:


contracts/cw-croncat/src/contract.rs
```diff
@@ -1,3 +1,5 @@
+//use std::ops::Add;
```
mikedotexe, line 1:
```suggestion
```

scripts/build-arm64.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Note: the below command has amd64 as the target, meaning M1 macs
+# Please use these directions if you're not on one:
+#   https://docs.cosmwasm.com/docs/1.0/getting-started/compile-contract#optimized-compilation
+  docker run --rm -v "$(pwd)":/code \
+  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+  --platform linux/arm64 \
+  cosmwasm/workspace-optimizer:0.12.6
```
mikedotexe, line 10: I think this version is good, as the docs also point to it. I'll note that we may want to be checking Docker Hub to see if there are newer versions.

https://hub.docker.com/r/cosmwasm/workspace-optimizer/tags

Looks like we're up to `0.12.8` now

mikedotexe (approved):

scripts/build-arm64.sh
deveusss: And new version reduces wasm size)

scripts/local/start-in-docker.sh
```diff
@@ -0,0 +1,208 @@
...
+
+
+
+#---------------------------------------------------------------------------
+echo "${Yellow}Instantiating smart contract...${NoColor}"
+IRES=$($BINARY tx wasm store /$DIR_NAME_SNAKE.wasm --from validator $TXFLAG --output json)
+CODE_ID=$(echo $IRES | jq -r '.logs[0].events[-1].attributes[0].value')
+echo "${Cyan}CODE_ID :" $CODE_ID "${NoColor}"
+
+INIT='{"denom":"'$STAKE'","cw_rules_addr":"$RULES_CONTRACT_ADDR"}'
```
NinoLipartiia, line 149:
```
INIT='{"denom":"'$STAKE'","cw_rules_addr":"'$RULES_CONTRACT_ADDR'"}'
```
It fixed the error with address validation for me

NinoLipartiia: Also I had to change `juno-node-1` to `juno_node_1`, not sure why it didn't work

mikedotexe:
I'm not sure which directory my terminal needs to be in order to run `scripts/local/start-in-docker.sh`

When I try going into `cw-croncat/contracts/cw-croncat` (where there is the contract's manifest `Cargo.toml`) I get:

Screen Shot 2022-09-21 at 11 08 00 AM

When I try to run it from `scripts/local` I get:

Screen Shot 2022-09-21 at 11 08 36 AM

When I try to run it from the project root I get:

Screen Shot 2022-09-21 at 11 09 08 AM

NinoLipartiia (approved):

PR #108, opened 2022-09-21
Calculate `agent_fee` as percentage of used gas

### LINKED ISSUE #112: Fix failure of `withdraw_reward` if agent received `agent_fee`
If agent calls `proxy_call` and doesn't execute a task, croncat accrues them small `agent_fee`, which agent can receive with `withdraw_reward`.  But since `agent_fee` isn't included in task balance, the contract doesn't have enough funds to pay agent, thus `withdraw_reward` fails.

I think we should revise payment logic to avoid errors like this.

NinoLipartiia:
Closes  #112
There are some changes of fee calculation:

- `agent_fee` is a percentage of gas (5%)
- address that creates a task pays for the gas + 5% of the gas
- if `proxy_call` fails the agent doesn't get reward and task balance doesn't change
- hence it fixes withdraw error #112


Buckram123:
Not sure about our business logic so shifting it to @mikedotexe and @TrevorJTClarke

TrevorJTClarke:
@Buckram123 @NinoLipartiia @mikedotexe the goal of agent rewards (taken from fees) will be accrued (not tokens sent every task), which accrue within the manager token holdings until the agent withdraws the tokens. This keeps the per-execution fees very low.

NinoLipartiia:
@TrevorJTClarke so now we accrue `agent_fee` reward (5ujunox) for every `proxy_call`. Agent accrues more ujunox if he executes a task. Do we need to send `agent_fee` in this case, or is `agent_fee` only a bonus for agent for calling `proxy_call` and not getting a task?

TrevorJTClarke:
> Do we need to send `agent_fee` in this case

We need to accrue fees for the agent, but not necessarily send every proxy_call.

Basic idea is this:
* Agent earns 5% of each task execution fee, above the full fee.
* Agent can withdraw to recoup the paid txn fee they paid to make sure the task executes
* the "5%" amount should be configured by manager owner (likely a DAO)

So in real amounts:
Task execution fee: 1000ujuno
Agent reward: 50ujuno
Agent accrued amount: 1050ujuno

mikedotexe:
Hey @NinoLipartiia is this PR ready for review? I wasn't sure if we chatted about anything in our latest meeting that would change the implementation, or if those changes might be coming in a future PR

NinoLipartiia:
Hey @mikedotexe, as I understand `agent_fee` should be different than just 5ujuno for every `proxy_call`, so I'll  work on it

TrevorJTClarke (approved):
Looks good! Thanks for making percent based in `calculate_required_amount `!

Was pretty tricky to review with all the extra merges in there.

contracts/cw-croncat/src/manager.rs
```diff
@@ -355,14 +314,27 @@ impl<'a> CwCroncat<'a> {
...
+        mut task: Task,
         queue_item: QueueItem,
     ) -> Result<Response, ContractError> {
         let task_hash = task.to_hash();
         // TODO: How can we compute gas & fees paid on this txn?
         // let out_of_funds = call_total_balance > task.total_deposit;
 
         let agent_id = queue_item.agent_id.unwrap();
+
+        if !queue_item.failed {
```
Buckram123, line 326: Hey @NinoLipartiia, pretty sure we want to `gas_fees` to the agent(because it's not not his fault and he used gas already for this task) even if this `queue_item` failed, wdyt?

PR #109, opened 2022-09-21
Fix `remove_task` to check owner for task with rules

NinoLipartiia:
Now we only check the owner for tasks without rules


contracts/cw-croncat/src/tasks.rs
```diff
@@ -477,6 +477,12 @@ impl<'a> CwCroncat<'a> {
                 .may_load(storage, &hash_vec)?
                 .ok_or(ContractError::NoTaskFound {})?;
 
+            if let Some(info) = info {
```
deveusss, line 480: Nice vulnerability fix, thank you)

deveusss (approved):

PR #110, opened 2022-09-22
Check balance for Staking task

NinoLipartiia:
I think that when `CosmosMsg::Staking` task is created, the balance for one execution should be gas + staking amount.

### DIFF
--- a/packages/cw-croncat-core/src/types.rs
+++ b/packages/cw-croncat-core/src/types.rs
@@ -376,8 +376,7 @@ impl Task {
                     if amount.amount.is_zero() {
                         valid = false;
                     }
-                    let add = vec![amount.clone()];
-                    amount_for_one_task.checked_add_native(&add)?;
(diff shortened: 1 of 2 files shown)


packages/cw-croncat-core/src/types.rs
```diff
@@ -367,6 +368,17 @@ impl Task {
...
+                CosmosMsg::Staking(StakingMsg::Delegate {
+                    validator: _,
+                    amount,
+                }) => {
+                    // Must attach enough balance for staking
+                    if amount.amount.is_zero() {
+                        valid = false;
+                    }
+                    let add = vec![amount.clone()];
+                    amount_for_one_task.checked_add_native(&add)?;
```
Buckram123, line 380:
```suggestion
                    amount_for_one_task.native.find_checked_add(amount)?;
```
redundant clone here

Buckram123 (approved):
Nice 👍

PR #118, opened 2022-09-25
don't double count native balances during instantiation

grapestems:
native balances are already included in the funds field during instantiation


deveusss (approved):
thank you)

PR #122, opened 2022-09-27
Removed duplicates and redundant clones, should reduce wasm size

### LINKED ISSUE #113: Address potential duplicate code in `proxy_call`
I noticed we have two blocks of code here:
https://github.com/CronCats/cw-croncat/blob/750199d4d23b253a0692ba3518a17083ab37623d/contracts/cw-croncat/src/manager.rs#L57-L77

that both contain:

```rs
        if some_hash.is_none() {
            let base_reward = self.send_base_agent_reward(deps.storage, agent, &info)?;
            return Ok(Response::new()
                .add_attribute("method", "proxy_call")
                .add_attribute("agent", &info.sender)
                .add_attribute("no_task_agent_base_reward", base_reward.to_string()));
        }
```

Feels like we'll want to take a look and simplify the logic

Buckram123:
Closes #113 
Also noticed some small changes for performance(like redundant clones)

Small bonus, `wasm` size:
722K->717K


deveusss (approved):

contracts/cw-croncat/src/manager.rs
```diff
@@ -451,23 +414,24 @@ impl<'a> CwCroncat<'a> {
...
-        let add_native = config.agent_fee;
-        agent.total_tasks_executed = agent.total_tasks_executed.saturating_add(1);
-        agent.balance.native.find_checked_add(&add_native)?;
-
-        // Reset missed slot
-        agent.last_missed_slot = 0;
-        self.agents.save(storage, &message.sender, &agent)?;
-        Ok(add_native)
-    }
+    /// TODO: Most likely unused but keep just in case
```
TrevorJTClarke, line 417: See lines 451-453, there is a case we need to make sure is covered:
```
/// Used in cases where there are empty slots or failed txns
/// Keep the agent profitable, as this will be a business expense
```

contracts/cw-croncat/src/manager.rs
```diff
@@ -1199,6 +1163,7 @@ mod tests {
             ("method", "remove_task"), // the last method
             ("ended_task", task_id_str.as_str().clone()),
         ];
+        println!("events: {:?}", res.events);
```
TrevorJTClarke, line 1166: should be removed?

contracts/cw-croncat/src/slots.rs
```diff
@@ -221,8 +223,8 @@ mod tests {
         // Empty slots
         store.time_slots.save(&mut deps.storage, 0, &vec![]).unwrap();
         store.block_slots.save(&mut deps.storage, 0, &vec![]).unwrap();
-        assert_eq!(None, store.pop_slot_item(&mut deps.storage, &0, &SlotType::Cron));
-        assert_eq!(None, store.pop_slot_item(&mut deps.storage, &0, &SlotType::Block));
+        assert_eq!(Err(ContractError::NoTaskFound{}), store.pop_slot_item(&mut deps.storage, 0, SlotType::Cron));
+        assert_eq!(Err(ContractError::NoTaskFound {  }), store.pop_slot_item(&mut deps.storage, 0, SlotType::Block));
```
TrevorJTClarke, line 227: this looks weird?

contracts/cw-croncat/src/state.rs
```diff
@@ -69,10 +69,10 @@ pub struct QueueItem {
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct TaskInfo {
-    pub task: Option<Task>,
+    pub task: Task,
     pub task_hash: Vec<u8>,
     pub task_is_extra: Option<bool>,
-    pub agent_id: Option<Addr>,
```
TrevorJTClarke, line 75: This should remain optional i believe

contracts/cw-croncat/src/tasks.rs
```diff
@@ -439,8 +439,7 @@ impl<'a> CwCroncat<'a> {
             for tid in time_ids {
                 let mut time_hashes = self.time_slots.may_load(storage, tid)?.unwrap_or_default();
                 if !time_hashes.is_empty() {
-                    time_hashes
-                        .retain(|h| String::from_utf8(h.to_vec()).unwrap() != task_hash.clone());
+                    time_hashes.retain(|h| h != &hash_vec);
```
TrevorJTClarke, line 442: ah, nice!

TrevorJTClarke (approved):
See minor comments, thanks for the cleanup here!

contracts/cw-croncat/src/state.rs
Buckram123: Are we planning to support non-agents complete tasks?

contracts/cw-croncat/src/slots.rs
Buckram123: Fixed!

contracts/cw-croncat/src/state.rs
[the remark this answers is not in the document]
Buckram123: This struct is used only inside `proxy_call()` and we allow calling it only for agents for now?
There is another struct for tasks, that we store in state, that doesn't have `agent_id` field
https://github.com/CronCats/cw-croncat/blob/b9ce3969dcba580cea33e66c48a1115e3f10e53f/packages/cw-croncat-core/src/types.rs#L209

PR #136, opened 2022-10-07
Add method for checking and deleting bad agents

### LINKED ISSUE #124: `last_missed_slot` is always zero
Field `last_missed_slot` in `Agent` should probably be updated by agent if they see the other agent miss a slot. Also we should use it to check the amount of tasks the agent skipped and eject bad agents.

NinoLipartiia:
Close #124

- added `last_executed_slot` for an agent, it is updated after every successful proxy call and stores this block number
- new method `Tick` unregisters agents which didn't act more than `cfg.agents_eject_threshold * cfg.slot_granularity` blocks

Also just noticed that in near implementation it's not allowed to delete the last agent, should we do the same here?


deveusss:
@NinoLipartiia need some fixes here)

deveusss (approved):
SGTM

PR #144, opened 2022-10-14
Tasks are grouped into slots according to their block height or timestamp

### LINKED ISSUE #138: Slot should consist of `slot_granularity` number of blocks
In current implementation a new slot is created for every block and `slot_granularity` parameter isn't used.

I think we need to implement something similar to this:
https://github.com/CronCats/contracts/blob/cafd3caafb91b45abb6e811ce0fa2819980d6f96/manager/src/lib.rs#L166-L187

NinoLipartiia:
Close #138 

Each block slot combines tasks from `slot_granularity_block` blocks, each time-based slot has tasks from some period of time (`slot_granularity_time`). Also implemented boundary validation for time-based tasks.


mikedotexe:
Hmm, one of the tests is failing with something that might be concerning @NinoLipartiia 

```
Caused by:
    0: serde parse error
       
       Caused by:
           missing field `id` at line 8 column 1
```

Thoughts?

deveusss:
@NinoLipartiia

deveusss (approved):

PR #126, opened 2022-09-29
Integrated tests, gas profiling and fees fixes

Buckram123:
Closes #125, fixes #117, resolves #83, resolves #82

PR #145, opened 2022-10-14
Follow up gas-benchmark

Buckram123:
-  Fixed fail on merge into master
-  Added calculation of average cost per action and base gas thanks for noticing @TrevorJTClarke  https://github.com/CronCats/cw-croncat/pull/126#discussion_r994922489
  And the values I'm currently getting: 
  Per non-wasm action: 124481.
  Base(proxy_call + unregister): 306762
  And for cw20 transfer action: 201279
- Added test of task with simple rule

As @mikedotexe mentioned in out last call, we can't predict gas consumption of random wasm actions, so I was thinking if we should encourage users to input `gas_limit` on those actions to not underrefund(or overrefund) agents. 
It would also be useful to add not turbo-absurd `gas_limit` default value for wasm actions here:
https://github.com/CronCats/cw-croncat/blob/af2f0969f01086f0f951985e61ce7da250bb2029/packages/cw-croncat-core/src/types.rs#L439-L443
Like double or tripple of `gas_action_fee`


mikedotexe:
hey @Buckram123 sorry I think other merges caused a conflict here, do you mind resolving?

deveusss (approved):

PR #157, opened 2022-10-19
Fleet/10 18 22

mikedotexe:
don't merge this, just passing the baton on some work, in case it's helpful


contracts/cw-croncat/src/manager.rs
```diff
@@ -304,10 +305,19 @@ impl<'a> CwCroncat<'a> {
...
         let (next_id, slot_kind) = task.interval.next(&env, task.boundary);
 
+        let recurring = task.interval == Interval::Once;
+        let enough_balance = task.verify_enough_balances(recurring);
+        if enough_balance.is_err() {
+            return Err(ContractError::CustomError {
+                val: format!("aloha {}", enough_balance.unwrap_err()),
+            });
+            // let hi = serde_json::from_str(&enough_balance.unwrap_err().to_string());
+        }
```
Buckram123, line 315: This is what it causes to fail, if task has not enough balance it falls and revert state here instead of unregistering the task

Buckram123 (changes requested):
Hey we good please look into #158!

contracts/cw-croncat/src/tasks.rs
```diff
@@ -91,7 +92,10 @@ impl<'a> CwCroncat<'a> {
                     .may_load(deps.storage, task_hash.as_bytes())?
             }
         };
-        Ok(res.map(Into::into))
+        if res.is_none() {
+            return Ok(None);
+        }
+        Ok(Some(TaskResponse::from(res.unwrap())))
```
Buckram123, line 98: Let's revert it https://doc.rust-lang.org/std/option/enum.Option.html#method.map !
equivalent of
```rust
res.map(Into::into)
```
would be
```rust
match res {
  Some(res) => Some(TaskResponse::from(res)), // same as Some(res.into()) or Some(Into::into(res))
  None => None
}
```
While the equivalent of
```rust
if res.is_none() {
  return Ok(None);
}
Ok(Some(TaskResponse::from(res.unwrap())))
```
would be
```rust
match res {
  None => return Ok(None),
  _ => ()
}
let res2 = match res {
  Some(res) => res,
  None => panic("called `Option::unwrap()` on a `None` value")
}
Ok(Some(TaskResponse::from(res2))))
```

Buckram123 (changes requested):
Hey, I guess rust having too many meaning to the keyword `map` :DD

mikedotexe:
Okay pretty sure we've discussed that we can close this since the useful, worthwhile parts have been incorporated in other PRs

mikedotexe:
I thought I closed this :)

PR #162, opened 2022-10-21
Remove task with rules when complete

mikedotexe:
It appears that we are not removing tasks with rules once they successfully complete.

Basically you call `proxy_call` with a task hash, which will check using the querier (you can search for `deps.querier.query_wasm_smart` in `proxy_call_with_rules`) and fail if the rules aren't met. If the rules _are_ met, it'll go the `proxy_callback` and inside there is a loaded conditional that looked like this:

```rs
if task.interval == Interval::Once
    || (task.stop_on_fail && queue_item.failed)
    || task.verify_enough_balances(false).is_err()
    // If the next interval comes back 0, then this task should not schedule again
    || next_id == 0
```

But those don't account for tasks with rules, so I added the condition.

I saw a test breaking so dove deeper into it and see that it was assuming a few things that didn't make sense to me. (I could be wrong, of course, hence reviewers)) )

### DIFF
--- a/contracts/cw-croncat/src/manager.rs
+++ b/contracts/cw-croncat/src/manager.rs
@@ -310,7 +310,6 @@ impl<'a> CwCroncat<'a> {
             || task.verify_enough_balances(false).is_err()
             // If the next interval comes back 0, then this task should not schedule again
             || next_id == 0
-            || task.with_rules()
         // proxy_call_with_rules makes it fail if rules aren't met
         {
             // Process task exit, if no future task can execute
--- a/contracts/cw-croncat/src/manager.rs
+++ b/contracts/cw-croncat/src/manager.rs
@@ -310,7 +310,8 @@ impl<'a> CwCroncat<'a> {
             || task.verify_enough_balances(false).is_err()
             // If the next interval comes back 0, then this task should not schedule again
             || next_id == 0
-            || task.with_rules() // proxy_call_with_rules makes it fail if rules aren't met
+            || task.with_rules()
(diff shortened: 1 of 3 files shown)


contracts/cw-croncat/src/manager.rs
```diff
@@ -310,6 +310,8 @@ impl<'a> CwCroncat<'a> {
             || task.verify_enough_balances(false).is_err()
             // If the next interval comes back 0, then this task should not schedule again
             || next_id == 0
+            || task.with_rules()
+        // proxy_call_with_rules makes it fail if rules aren't met
```
Buckram123, line 314: Task with rules can't be recurring?

Buckram123 (commented):
I can imagine tasks with rules can be recurring. You sure we don't remove them same way we do with the tasks without rules?

contracts/cw-croncat/src/manager.rs
mikedotexe: good question

mikedotexe: let's see… so where else would this removal happen? will dig through the code more

mikedotexe: No i think you're right. I think I was confused because the tests seemed to think it should be gone, but perhaps it was gone just because with the old code it was so expensive, it would fail because it ran out of balance.

AKA this condition (a little above where you highlighted here):

`|| task.verify_enough_balances(false).is_err()`

PR #176, opened 2022-10-25
Stacking rules, Injectable messages

### LINKED ISSUE #64: CW-RULES: Generic MSG deserialize/serialize + stacking
GOAL:
Allow multiple rules to stack for enabling very complex logical paths pre-execution. This is only possible IF the deserialization is possible on-chain.
The stacking of query messages and their internal variables is key here. If a rule does not need previous arguments, then it can simply be evaluated in a serial fashion. If a rule needs variables computed from previous query, then stacking with input enables a vast amount of other possibilities.

Example use case:
`If the prices of NFT is above XX price, and I am NOT the owner of NFT, place a bid/buy on the NFT.`

Which breaks down to:
1. `query price of NFT`
2. `use price to check if above XX in another helper method in cw-rules`
3. `query owner of NFT`
4. `EXEC buy msg`

Basic data structures:
```
// Request:
{
  contract_address: "",
  msg: ""
}
```

```
// pub type RuleResponse<T> = (bool, T);
(true, "some_parsable_format_here")
```

Critical Items to research:
- [ ] non-float deserialize for msg matching, using the second piece of tuple RuleResponse as the parsed data
- [ ] msg serialization with unknown msg structs

Requirements:
- [ ] Using branch: https://github.com/CronCats/cw-croncat/pull/60
- [ ] Add support for generic rules, looping through each rule and then stacking previous response if any
- [ ] "Stack" each next query (first one doesnt need, as we dont have data yet), where the previous second piece of tuple RuleResponse should get injected as a variable within the next rule msg
- [ ] Final outcome of this query function should return RuleResponse where the `T` is the output of the LAST rule output

Buckram123:
Closes #64

- [x]  Implementation and tests inside `cw-rules` contract
- [x] Test from `cw-croncat` side


mikedotexe (changes requested):
hey @Buckram123 can we take a look at the linting here pls:

Screen Shot 2022-10-31 at 1 32 04 PM

And a minor merge conflict 🙏🏼

Buckram123:
Implementing #189 here, I need stacking rules for this

contracts/cw-croncat/src/error.rs
```diff
@@ -36,9 +36,12 @@ pub enum ContractError {
     #[error("Rules are not ready. Failed at rule {index:?}")]
     RulesNotReady { index: u64 },
 
-    #[error("No rules for this task hash")]
+    #[error("No rules for this task hash: {task_hash}")]
```
mikedotexe, line 39: great catch

contracts/cw-rules/Cargo.toml
```diff
@@ -32,8 +32,9 @@ optimize = """docker run --rm -v "$(pwd)":/code \
 
 [dependencies]
 cw-croncat-core = { version = "0.1.3", path = "../../packages/cw-croncat-core" }
-generic-query = { version = "0.1.0", path = "../../packages/generic-query" }
-cw-rules-core = { version = "0.1.1", path = "../../packages/cw-rules-core" }
+generic-query = { version = "0.1.1", path = "../../packages/generic-query" }
+smart-query = { version = "0.1.0", path = "../../packages/smart-query" }
+cw-rules-core = { version = "0.1.2", path = "../../packages/cw-rules-core" }
```
mikedotexe, line 37: Thank you for being diligent in bumping versions

mikedotexe:
hey @Buckram123 looks like this was moved out of draft? I know this is a big one, feel free to let us know when we can re-review and the merge conflicts are good 🙏🏼

Buckram123:
Currently blocked by wasm-size, have few ideas where I can decrease it, hope it will get through before splitting contract

Buckram123:
Currently blocked by wasm-size, have few ideas where I can decrease it, hope it will get through before splitting contract

mikedotexe:
> Currently blocked by wasm-size, have few ideas where I can decrease it, hope it will get through before splitting contract

Gotcha, thanks! Also, if you get to a point where you think it's worth talking about splitting up the contract now, that's a totally reasonable conversation to have. Maybe we can talk about this as a team soon. Thanks so much for the effort on this, it's seriously exciting. I found myself pacing around my house thinking about how cool this is going to be

Buckram123:
update: Heyyy, after few small changes still had to sacrifice query non-critical(likely even redundant) information, but now unblocked, so testing locally and on-chain into opening this PR

contracts/cw-croncat/src/tests/helpers.rs
```diff
@@ -327,12 +331,13 @@ pub fn default_task() -> Task {
             start: None,
             end: None,
         },
-        funds_withdrawn_recurring: Default::default(),
+        //funds_withdrawn_recurring: Default::default(),
```
mikedotexe, line 334: Hey, curious about this part? Are we removing these to reduce the contract size a bit?
I was trying this out locally and my indexer broke because it's expecting to store values from here.

mikedotexe (commented):
Would love to chat about this during tomorrow's meeting. This might be one of the most interesting PRs we've had in months.

I left another comment about the indexer crashing, expecting a field, but I'll explain the reason why I wanted to run the indexer on it.. So, I was doing basic tests after deploying, instantiating, creating a task, starting the agent, etc.. And I'm pretty sure the same recurring task that I've been doing for a while (simple payroll to three people) completed much earlier than I expected. I noticed the agent logs being uninteresting (no big blocks of logs, like when it's actively doing payroll) and did a query to show the tasks, and saw there were none. So it must have completed and refunded to the rest to the task creator.

I suspect we need to check the gas on this carefully, and compare it to the `main` branch. Perhaps you can try running it and seeing if you see the same thing.

Here's a messy, but simple payroll bash script. (I'm not even using env vars, just pasting the new contract address every time)

[recurring-payroll.sh.removethispartcuzgithub.log](https://github.com/CronCats/cw-croncat/files/10109829/recurring-payroll.sh.removethispartcuzgithub.log)

contracts/cw-croncat/src/tests/helpers.rs
Buckram123: Yeah, needed last few kb to cut :D, here is the comment tho:
https://github.com/CronCats/cw-croncat/blob/6fda939ac75834938775fe59137bf46d47637963/packages/cw-croncat-core/src/types.rs#L317-L318

mikedotexe: Awesome! Did I understand from today's meeting that you'll remove the field altogether? Making sure

Buckram123: Done)

contracts/cw-croncat/src/agent.rs
```diff
@@ -101,14 +101,16 @@ impl<'a> CwCroncat<'a> {
...
+        self.balancer
+            .get_agent_tasks(
+                &deps,
+                &env,
+                &self.config,
+                &self.agent_active_queue,
+                account_id,
+                (Some(block_slots as u64), Some(time_slots as u64)),
+            )
+            .map_err(|err| StdError::generic_err(err.to_string()))
```
mikedotexe, line 113: good eye

contracts/cw-croncat/src/agent.rs
```diff
@@ -333,7 +335,13 @@ impl<'a> CwCroncat<'a> {
             let kicked_agents: Vec<Addr> = {
                 let mut kicked = Vec::with_capacity(agent_position);
                 for _ in 0..=agent_position {
-                    kicked.push(self.agent_pending_queue.pop_front(deps.storage)?.unwrap());
+                    let agent = self.agent_pending_queue.pop_front(deps.storage)?;
+                    // Since we already iterated over it - we know it exists
+                    let kicked_agent;
+                    unsafe {
```
mikedotexe, line 341: interested in hearing about this `unsafe` usage here 🍿

contracts/cw-croncat/src/owner.rs
```diff
@@ -349,27 +345,27 @@ impl<'a> CwCroncat<'a> {
...
-                        action_idx: item.action_idx.into(),
-                        task_hash: item.task_hash,
-                        task_is_extra: item.task_is_extra,
-                        agent_id: item.agent_id,
-                        failed: item.failure.is_some(),
-                    },
-                }
-            })
-            .collect();
+        // let reply_queue: Vec<ReplyQueueResponse> = self
```
mikedotexe, line 348: I know this is a pretty big PR, so wanted to make sure we intended to keep these commented out. (IMO commenting out is actually fine at this stage, just want to call it out in case it's helpful and you're like, "oh wait, no we can bring that back actually")

Buckram123: I guess it's fine to keep this field for debugging later, reply queue should always be empty, when we query it, because we clean it up at the and of every proxy_call(also there shouldn't be more than 1 Item too).

contracts/cw-croncat/src/agent.rs
Buckram123: If we already found agent at N position, we can safely assume that there is at least N agents, so we can safely unwrap an option, and we don't need check, so IMO it's perfect place for `unwrap_unchecked()`

Buckram123:
DEMO: https://github.com/Buckram123/cw-croncat/pull/1

packages/cw-croncat-core/src/msg.rs
```diff
@@ -319,7 +320,9 @@ pub struct CwCroncatResponse {
     pub time_slots_rules: Vec<SlotWithRuleResponse>,
     pub block_slots_rules: Vec<SlotWithRuleResponse>,
 
-    pub reply_queue: Vec<ReplyQueueResponse>,
+    /// Pretty sure it should be always empty
+    /// P.S: we can recover it after split, needed few non-critical Kbs
+    // pub reply_queue: Vec<ReplyQueueResponse>,
```
TrevorJTClarke, line 325: im in agreement, let's remove and remove the comments

TrevorJTClarke (approved):
love the huggeeeeeee progress on this!! Let's get this out in testnet for further feedback loops :D

mikedotexe (approved):
Thank you for the demo here, too:
https://github.com/Buckram123/cw-croncat/blob/2cec3d4b9535d28839c47f463679f099f7b731f4/demo/DEMO.md

TrevorJTClarke (approved):
HERE WE GOOOOOOOOOOO!!!!!

PR #179, opened 2022-10-26
feat: Task version

### LINKED ISSUE #161: Add version to each task
Make sure we add a version to each task when it's created. Let's use the same logic that sets the contract version.

`const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");`

I bet there will be a little bit of fixes needed for the unit tests, too))

deveusss:
Closes #161


contracts/cw-croncat/src/tasks.rs
```diff
@@ -269,6 +269,7 @@ impl<'a> CwCroncat<'a> {
             .native
             .find_checked_add(&coin(price, &cfg.native_denom))?;
 
+        const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
```
Buckram123, line 272: I think we can reuse `CONTRACT_VERSION` from here: https://github.com/CronCats/cw-croncat/blob/c43946cfafe9aca4e519173c86beddc85b58f8c4/contracts/cw-croncat/src/contract.rs#L15

Buckram123 (approved):
Nice one

contracts/cw-croncat/src/tests/helpers.rs
```diff
@@ -180,8 +180,12 @@ pub fn add_1000_blocks(block: &mut BlockInfo) {
     block.time = block.time.plus_seconds(10);
     block.height += 1000;
 }
-
+pub fn get_contract_version() -> String {
```
mikedotexe, line 183: Reading the README of the `cw2` contract (the one we use in the `instantiate` endpoint) to set the contract version, it looks like we already have a preferred way to get the contract verison:
https://crates.io/crates/cw2/0.16.0

> You can query using:
> 
>     wasmd query wasm contract-state raw [contract_addr] 636F6E74726163745F696E666F --node $RPC

mikedotexe:
We've been having good discussion about this in our chat, but wanted to fill in some details.
We've established that the cw2 dependency is already saving our contract's version, so we can use that and remove the function added in this PR named: `get_contract_version`.

You can see this in action here:

Screen Shot 2022-10-31 at 2 10 48 PM

That is doing a query to a specific key prefix that is saved on our contract already.

After we remove that function from this PR, we can modify the tests so they're able to check the version of the contract.

@deveusss you linked to a great resource, which is the `lib.rs` in the cw2 crate:

https://github.com/CosmWasm/cw-plus/blob/44c06c9fa0ecd5f8e61b156d7e726df2a4c0d66d/packages/cw2/src/lib.rs#L66-L70

We can use that, except we can make it a little easier, where the `key` given in the link above can be replaced with this variable:

`let new_key: cosmwasm_std::Binary = [99, 111, 110, 116, 114, 97, 99, 116, 95, 105, 110, 102, 111].into();`

mikedotexe:
hey @deveusss do you think we should go with this approach for now? Don't want to merge if you're in the middle of doing the changes talked about in https://github.com/CronCats/cw-croncat/pull/179#issuecomment-1297719556

deveusss:
@yes, will change the PR first then will request for rereview) thank you @mikedotexe )

deveusss:
Done

packages/cw-croncat-core/src/query.rs
```diff
@@ -0,0 +1,21 @@
...
+
+pub struct CroncatQuerier<'a> {
+    querier: &'a QuerierWrapper<'a>,
+}
+impl<'a> CroncatQuerier<'a> {
+    pub fn new(querier: &'a QuerierWrapper<'a>) -> Self {
+        CroncatQuerier { querier }
+    }
+
+    pub fn query_contract_info(&self, contract_address: String) -> StdResult<ContractVersion> {
```
mikedotexe, line 12: ❤️

mikedotexe (approved):
thank you!

PR #182, opened 2022-10-31
feat: Nominate first agent in pending queue if last active agent got unregistered or kicked

### LINKED ISSUE #181: If active agents is empty, the first pending agent should have Nominated status
For this ticket, we can use this Juno testnet address for the Manager contract:

`juno1cz5ru2px6cr4l27rxyk7md7976ezgeh8tu3mumhwp3ua2wspcfxszss3mr`

Clone the agent:
https://github.com/CronCats/croncat-rs

If you don't have two agents registered, make sure to create both with the commands:

    cargo run generate-mnemonic
    cargo run generate-mnemonic --new-name aloha

In the agent's codebase, change the value of `contract_address` to be the one above, in `config.uni-5.yaml`.

Then run:

    cargo run register-agent --chain-id uni-5

This will read the contract from the `config.uni-5.yaml` and attempt to register. It should succeed and look like this:

Screen Shot 2022-10-26 at 6 59 38 PM

You can check the agents on the contract using the following `junod` command:

    junod q wasm contract-state smart juno1cz5ru2px6cr4l27rxyk7md7976ezgeh8tu3mumhwp3ua2wspcfxszss3mr '{"get_agent_ids":{}}' --node https://juno-testnet-rpc.polkachu.com:443 --chain-id uni-5 -o json | jq

(Make sure you have installed or install [`jq`](https://stedolan.github.io/jq/))

At this point, it should return something like this:

```json
{
  "data": {
    "active": [
      "juno1hntaw5cgvxz0gftpd9gec876n0l6az20clrred"
    ],
    "pending": []
  }
}
```

Next, let's register the second agent:

    cargo run register-agent --chain-id uni-5 --sender-name aloha

Now both will be there when you run the `juno` query command, one in the active, one in the pending. Now let's have the active agent unregister

    cargo run unregister-agent --chain-id uni-5

If you run the `junod` query again, you'll see there are no active agents, and there is one agent in the pending queue. This agent should have a status of "Nominated" because we want them to enter the active queue.

You can see, however, that the status is Pending. Something is wrong with that logic there. You can query to see the agent details including status with a command like this:

    junod q wasm contract-state smart juno1cz5ru2px6cr4l27rxyk7md7976ezgeh8tu3mumhwp3ua2wspcfxszss3mr '{"get_agent":{"account_id":"juno1hntaw5cgvxz0gftpd9gec876n0l6az20clrred"}}' --node https://juno-testnet-rpc.polkachu.com:443 --chain-id uni-5 -o json | jq

You'll see something like this:

```
{
  "data": {
    "status": "Pending",
    "payable_account_id": "juno1hntaw5cgvxz0gftpd9gec876n0l6az20clrred",
    "balance": {
      "native": [],
      "cw20": []
    },
    "total_tasks_executed": 0,
    "last_executed_slot": 505822,
    "register_start": "1666836892900035336"
  }
}
```

Buckram123:
Closes #181

### DIFF
--- a/contracts/cw-croncat/src/agent.rs
+++ b/contracts/cw-croncat/src/agent.rs
@@ -24,35 +24,27 @@ impl<'a> CwCroncat<'a> {
     ) -> StdResult<Option<AgentResponse>> {
         let account_id = deps.api.addr_validate(&account_id)?;
         let agent = self.agents.may_load(deps.storage, &account_id)?;
-        if agent.is_none() {
+        let a = if let Some(a) = agent {
+            a
+        } else {
             return Ok(None);
-        }
-        let active: Vec<Addr> = self.agent_active_queue.load(deps.storage)?;
-        let a = agent.unwrap();
-        let mut agent_response = AgentResponse {
-            status: AgentStatus::Pending, // Simple default
+        };
+
+        let agent_status = self
+            .get_agent_status(deps.storage, env, account_id)
+            // Return wrapped error if there was a problem
+            .map_err(|err| StdError::GenericErr {
+                msg: err.to_string(),
+            })?;
+
+        let agent_response = AgentResponse {
+            status: agent_status,
             payable_account_id: a.payable_account_id,
             balance: a.balance,
             total_tasks_executed: a.total_tasks_executed,
             last_executed_slot: a.last_executed_slot,
             register_start: a.register_start,
         };
-
-        if active.contains(&account_id) {
-            agent_response.status = AgentStatus::Active;
-            return Ok(Some(agent_response));
-        }
-
-        let agent_status = self.get_agent_status(deps.storage, env, account_id);
-
-        // Return wrapped error if there was a problem
-        if agent_status.is_err() {
-            return Err(StdError::GenericErr {
-                msg: agent_status.err().unwrap().to_string(),
-            });
-        }
-
-        agent_response.status = agent_status.expect("Should have valid agent status");
         Ok(Some(agent_response))
     }
 
@@ -291,6 +283,7 @@ impl<'a> CwCroncat<'a> {
         // Compare current time and Config's agent_nomination_begin_time to see if agent can join
         let c: Config = self.config.load(deps.storage)?;
 
+        let active_agents: Vec<Addr> = self.agent_active_queue.load(deps.storage)?;
         let time_difference =
             if let Some(nomination_start) = self.agent_nomination_begin_time.load(deps.storage)? {
                 env.block.time.seconds() - nomination_start.seconds()
@@ -301,9 +294,9 @@ impl<'a> CwCroncat<'a> {
                 });
             };
         // Agent must be in the pending queue
-        let pending_queue = self.agent_pending_queue.load(deps.storage)?;
+        let pending_queue: Vec<Addr> = self.agent_pending_queue.load(deps.storage)?;
         // Get the position in the pending queue
-        if let Some(agent_position) = pending_queue
+        let kicked_agents = if let Some(agent_position) = pending_queue
             .iter()
             .position(|address| address == &info.sender)
         {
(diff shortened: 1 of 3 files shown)


contracts/cw-croncat/src/agent.rs
```diff
@@ -316,23 +309,20 @@ impl<'a> CwCroncat<'a> {
...
-                }
+                let mut pending_queue = pending_queue;
+                let kicked_agents: Vec<Addr> = pending_queue.drain(..=agent_position).collect();
                 self.agent_pending_queue
-                    .save(deps.storage, &pending_agents)?;
+                    .save(deps.storage, &pending_queue)?;
 
                 // and adding to active queue
-                let mut active_agents: Vec<Addr> = self.agent_active_queue.load(deps.storage)?;
+                let mut active_agents = active_agents;
```
mikedotexe, line 318: This makes it mutable?

contracts/cw-croncat/src/helpers.rs
```diff
@@ -127,59 +127,65 @@ impl<'a> CwCroncat<'a> {
...
+        // Then determine if this agent is considered "Nominated" and should call CheckInAgent
+        let max_agent_index = self.max_agent_index(storage, &c, env, &(active.len() as u64))?;
+        let agent_status = match max_agent_index {
+            Some(max_idx) if agent_position as u64 <= max_idx => AgentStatus::Nominated,
+            _ => AgentStatus::Pending,
+        };
+        Ok(agent_status)
+    }
+
+    pub(crate) fn max_agent_index(
```
mikedotexe, line 157: Could we add a triple-backslash comment here talking about what this method does? 🙏🏼

mikedotexe:
Hey, thanks for doing this. Sorry, but can we rework this so the first pending agent isn't automatically put into the active queue? We'd still expect agents in the pending queue to be running, and don't want to promote an agent that might not be running anymore. Running agents will be checking to see if they have the `Nominated` status, and as you remember, as time goes on, the next pending agents are allowed to get in first if they're not online and checking.

mikedotexe (changes requested):
Hey, thanks for doing this. Sorry, but can we rework this so the first pending agent isn't automatically put into the active queue? We'd still expect agents in the pending queue to be running, and don't want to promote an agent that might not be running anymore. Running agents will be checking to see if they have the `Nominated` status, and as you remember, as time goes on, the next pending agents are allowed to get in first if they're not online and checking.

contracts/cw-croncat/src/agent.rs
Buckram123: yeah, we can play with mutability of variable if we own it any time
```rust
  // not mutable
  let a = 5;
  // mutable, quick change in close scope
  let mut a = a;
  a = 6;
  // not mutable
  let a = a;
```

Buckram123:
> We'd still expect agents in the pending queue to be running, and don't want to promote an agent that might not be running anymore.

If first agent from pending queue is offline and we have tasks to do, isn't it okay to kick him for being not active and move on until we find online agent?

> Running agents will be checking to see if they have the `Nominated` status, and as you remember, as time goes on, the next pending agents are allowed to get in first if they're not online and checking.

I kinda missed the point, right now the timer starts as soon as new task is created, do we want to make first pending agent `Nominated` if active queue is empty even if timer haven't been started yet? @mikedotexe

mikedotexe:
> I kinda missed the point, right now the timer starts as soon as new task is created, do we want to make first pending agent `Nominated` if active queue is empty even if timer haven't been started yet? @mikedotexe

Let's see… so my understanding is that in the CronCat Manager Config, we have `agent_nomination_begin_time` which gets set every time a new task is added, here:

(this is in the `create_task` function)
https://github.com/CronCats/cw-croncat/blob/8c85201856c3dfa89069b5fe97540c3f0d5ee5fa/contracts/cw-croncat/src/tasks.rs#L373-L374

Maybe we can walk through the scenario you're imagining?

Also, I'm wondering if you're suggesting that right here:
https://github.com/CronCats/cw-croncat/blob/8c85201856c3dfa89069b5fe97540c3f0d5ee5fa/contracts/cw-croncat/src/helpers.rs#L165-L166

We check to see if there's an empty  active queue, and if there is, immediately return `AgentStatus::Nominated`?

Buckram123:
> We check to see if there's an empty active queue, and if there is, immediately return `AgentStatus::Nominated`?

Yeah, but only for first agent, for the rest of the agents nomination works as normal

mikedotexe:
> Yeah, but only for first agent, for the rest of the agents nomination works as normal

Okay, yes, I agree that's good logic to add, thank you!

contracts/cw-croncat/src/helpers.rs
```diff
@@ -127,59 +127,72 @@ impl<'a> CwCroncat<'a> {
...
+            };
 
-            // If we should allow a new agent to take over
-            let num_agents_to_accept =
-                self.agents_to_let_in(&min_tasks_per_agent, &num_active_agents, &total_tasks);
-            let agent_nomination_begin_time = self.agent_nomination_begin_time.load(storage)?;
-            match agent_nomination_begin_time {
-                Some(begin_time) if num_agents_to_accept > 0 => {
+        // Edge case if last agent unregistered
+        if active.is_empty() && agent_position == 0 {
```
mikedotexe, line 148: Oh i see you added it here 👍🏼

contracts/cw-croncat/src/tests/agent.rs
```diff
@@ -924,3 +924,89 @@ fn test_query_get_agent_tasks() {
         .into()
     );
 }
+
+#[test]
+fn test_last_unregistered_active_agent_promotes_first_pending() {
```
mikedotexe, line 929: Thank you for adding this test!

mikedotexe (approved):

PR #185, opened 2022-11-03
Migrate pending queue to deque

### LINKED ISSUE #183: Migrate agent queues from `Vec` to `VecDeque`
If `agent_pending_queue` will be too huge in the future we might fail to get it from storage, so my plan is to move from `Vec` to `cw-storage-plus` `Deque`:
https://github.com/CosmWasm/cw-storage-plus/blob/main/README.md#vecdeque

Another random benefit: it will be cheaper to pop from the start, and we do that every time when agent gets promoted to the active queue and poped from the start of the pending queue

Buckram123:
Closes #183


contracts/cw-croncat/src/owner.rs
```diff
@@ -403,7 +403,11 @@ impl<'a> CwCroncat<'a> {
             config: self.query_config(deps)?,
 
             agent_active_queue: self.agent_active_queue.load(deps.storage)?,
-            agent_pending_queue: self.agent_pending_queue.load(deps.storage)?,
+            agent_pending_queue: self
+                .agent_pending_queue
+                .iter(deps.storage)?
+                .take(50)
```
mikedotexe, line 409: I think this will work for now, eventually we'd like to have pagination, but don't feel it's going to get past 50 pending agents for a while, and likely not critical to know about all of them anyway

contracts/cw-croncat/src/state.rs
```diff
@@ -101,7 +101,7 @@ pub struct CwCroncat<'a> {
     pub agents: Map<'a, &'a Addr, Agent>,
     // TODO: Assess if diff store structure is needed for these:
     pub agent_active_queue: Item<'a, Vec<Addr>>,
-    pub agent_pending_queue: Item<'a, Vec<Addr>>,
+    pub agent_pending_queue: Deque<'a, Addr>,
```
mikedotexe, line 104: Good find with the Deque!

contracts/cw-croncat/src/tests/agent.rs
```diff
@@ -1008,3 +1008,140 @@ fn test_last_unregistered_active_agent_promotes_first_pending() {
         }
     );
 }
+
+#[test]
+fn removing_agent_from_any_side_is_working() {
```
mikedotexe, line 1013: Again, really appreciate you adding tests with these

contracts/cw-croncat/src/agent.rs
```diff
@@ -355,10 +360,12 @@ impl<'a> CwCroncat<'a> {
 
     /// Removes the agent from the active set of agents.
     /// Withdraws all reward balances to the agent payable account id.
+    /// In case it fails to unregister pending agent try to set `from_behind` to true
```
mikedotexe, line 363: @Buckram123 would you mind taking a little time to help me understand the `from_behind` idea? If it's easy to do here, that's great, or if it's better to explain on our next call, that works, too

mikedotexe (changes requested):
(just that explanation of `from_behind` since I'm not seeming to understand it yet)

contracts/cw-croncat/src/agent.rs
Buckram123: Hey, thanks for asking!
In short: there is a slight chance, that removing an agent could be impossible from left to right, so agent could try to remove himself from right to left.
In shorter: Gas.

contracts/cw-croncat/src/owner.rs
Buckram123: Honestly, I would like to separate query of pending/active agents queues from this method, to have an easier time for pagination

mikedotexe (approved):

Issue #134, opened 2022-10-06
[FEATURE] CI/CD pipeline


TrevorJTClarke:
### GOAL:
Add github configurations to create reproducible builds of all contracts after completing tests, linting, etc.

### Requirements:
- [ ] Continue based on PR: https://github.com/CronCats/cw-croncat/pull/126
- [ ] Setup CI flows for:
  - [x] [Audit](https://github.com/DA0-DA0/dao-contracts/blob/main/.github/workflows/audit.yml)
  - [x] [Linting/Schema](https://github.com/DA0-DA0/dao-contracts/blob/main/.github/workflows/basic.yml)
  - [x] [Code coverage](https://github.com/DA0-DA0/dao-contracts/blob/main/.github/workflows/codecov.yml)
  - [ ] [E2E Tests](https://github.com/DA0-DA0/dao-contracts/blob/main/.github/workflows/integration_tests.yml)
  - [x] [exported types](https://github.com/DA0-DA0/dao-contracts/blob/main/.github/workflows/ts_types.yml)
- [x] Output of contracts should be reproducible `.wasm` files
- [x] Store the checksum of all `.wasm` files, for use within upcoming factory registry in a text file (or similar)

PR #201, opened 2022-11-29
Checksum action

Buckram123:
Part of #134


TrevorJTClarke (approved):
YASSSSSS!!!!
Can you make sure the gas check passes? then GTG

PR #203, opened 2022-12-02
fix: Active agent error message fix

deveusss:
closes #202

### DIFF
--- a/contracts/cw-croncat/src/tests/agent.rs
+++ b/contracts/cw-croncat/src/tests/agent.rs
@@ -917,7 +917,8 @@ fn test_query_get_agent_tasks() {
     assert_eq!(
         query_task_res.unwrap_err(),
         cosmwasm_std::StdError::GenericErr {
-            msg: "Querier contract error: Generic error: Agent not registered".to_string()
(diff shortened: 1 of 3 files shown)


contracts/cw-croncat/src/tests/agent.rs
```diff
@@ -917,7 +917,8 @@ fn test_query_get_agent_tasks() {
     assert_eq!(
         query_task_res.unwrap_err(),
         cosmwasm_std::StdError::GenericErr {
-            msg: "Querier contract error: Generic error: Agent not registered".to_string()
+            msg: "Querier contract error: Generic error: Agent is not in the list of active agents"
```
TrevorJTClarke, line 920: why do we have this `Querier contract error: Generic error:` seems redundant?

TrevorJTClarke (approved):
Looks great!! minor tiny comment

contracts/cw-croncat/src/tests/agent.rs
deveusss: In queries we return   return Err(StdError::GenericErr {
                msg: AgentNotRegistered {}.to_string(),
            });

PR #208, opened 2022-12-09
[FIX] Get rid of `GenericBalance` and `GetConfigResponse`. Move `agent_active_indices` from `Config`

Buckram123:
- [x] Contract balances
- [x] Agents balances
- [x] Tasks balances
- [x] Remove `GenericBalance` traits and struct from the ecode


Buckram123:
Oops, tried to use `HashMap` for tasks balance, thought [this one](https://github.com/CosmWasm/serde-json-wasm/pull/45) is stable. Is it ok to use beta version: https://github.com/CosmWasm/cosmwasm/releases/tag/v1.2.0-beta.0
@TrevorJTClarke

Buckram123:
Forget about my latest comment, I think we might want to use this struct instead:
```rust
TaskBalance {
  native: Vec<Coin> // can contain IBC coins in tx
  cw20: Cw20CoinVerified // Up to one cw20 coin in task to not have too suspicious txs
}
```
@TrevorJTClarke thoughts? Can you think of use case of multiple `cw20` in a single task?

TrevorJTClarke:
> Can you think of use case of multiple `cw20` in a single task?

I can - so we need to decide if we support it or not :)
So, if you have 3 actions, you could potentially be doing 3 different cw20 transfers.

Buckram123:
> I can - so we need to decide if we support it or not :) 

It's really tough, wonder if it can create any vulnerability. Don't think so, have nothing against supporting multiple `cw20`. 

@deveusss What do you think?

deveusss:
> cw20 transfers.

Why not use standard balance or WrappedBalance(Balance) so that only one type of coin/cw20 could exist in actions.
For transferring another coin/cw20 new task should be created, this will keep our tasks clean and consistent with CW core functionality.

TrevorJTClarke:
> > cw20 transfers.
> 
> Why not use standard balance or WrappedBalance(Balance) so that only one type of coin/cw20 could exist in actions. For transferring another coin/cw20 new task should be created, this will keep our tasks clean and consistent with CW core functionality.
> 
> @TrevorJTClarke @Buckram123

That sounds good to me! Simpler the safer here!

contracts/cw-croncat/src/owner.rs
```diff
@@ -31,18 +29,48 @@ impl<'a> CwCroncat<'a> {
...
-            staked_balance: c.staked_balance,
             limit: c.limit,
         })
     }
 
-    pub(crate) fn query_balances(&self, deps: Deps) -> StdResult<GetBalancesResponse> {
+    pub(crate) fn query_balances(
+        &self,
+        deps: Deps,
+        from_index: Option<u64>,
```
mikedotexe, line 39: Awesome, thanks for including pagination

mikedotexe (approved):
This is looking great

contracts/cw-croncat/src/agent.rs
```diff
@@ -240,21 +237,43 @@ impl<'a> CwCroncat<'a> {
...
+
+        let (messages, native, cw20) = self.agent_withdraw_messages(
+            storage,
+            agent_id,
+            &agent.payable_account_id,
+            native_keys,
+            cw20_keys,
+        )?;
+        for coin in native {
+            self.sub_availible_native(storage, &coin)?;
```
mikedotexe, line 271: We have a few places with "available" misspelled :)

deveusss: thx Mike, good eye)

Buckram123: > We have a few places with "available" misspelled :)

Oops, mybad! I totally should download some spellchecker for my IDE

Buckram123:
Closes #239, #240

deveusss:
@Buckram123 fix please this one, so we can merge?

Buckram123:
> @Buckram123 fix please this one, so we can merge?

Done!)

deveusss:
@Buckram123

deveusss:
hey @Buckram123 are this changes moved to new split?

Buckram123:
moved to the #250

PR #214, opened 2022-12-14
[RESEARCH] of gas consumption during cross contract communication

Buckram123:
Compares:
- Simple Function call **VS** Function call from "stateless" library contract
- Raw state query **VS** Smart query of the external contract
- Function call of "stateless" library contract **VS** Function call of "stateless" library contract with `reply()`


Buckram123:
Added:

- Execute contract **VS** Execute contract with reply
- Query same contract twice **VS** Query two different contracts
- Execute same contract twice **VS** Execute two different contracts

TrevorJTClarke:
Hey @Buckram123 whats the goal with this PR?
Im wondering if this would be better suited outside this repo. Perhaps in a "research" repo or within cw-purrbox.
Thoughts?

Buckram123:
> Hey @Buckram123 whats the goal with this PR? Im wondering if this would be better suited outside this repo. Perhaps in a "research" repo or within cw-purrbox. Thoughts?

PR #209, opened 2022-12-10
[FEATURE] Add "simulate" query msg for estimating gas & occurrences

deveusss:
closes #204

PR #216, opened 2022-12-14
[FIX]: create_task response data

### LINKED ISSUE #207: Adding useful reply data to `create_task`
Aim: Increase usability of `create_task` msg

(This might be useful for other messages)

At the moment, to use cw-croncat to `create_task` we rely on the reply attributes in the wasm event to get the `task_hash` to [update our state](https://github.com/nymlab/vectis-plugins/blob/8215707eee771f8dccad5ded3f688fb8bd5e8b6c/cronkitty/src/lib.rs#L53). 

Perhaps it can be return in the `data` field of the `Response` so it can be easily parsed.

deveusss:
closes #207


contracts/cw-croncat/src/tasks.rs
```diff
@@ -411,6 +411,7 @@ impl<'a> CwCroncat<'a> {
         };
 
         Ok(Response::new()
+            .set_data(format!("task_hash {}", hash).into_bytes())
```
Buckram123, line 414:
```suggestion
            .set_data(hash.as_bytes())
```
I think should work

Buckram123 (changes requested):
small fix

contracts/cw-croncat/src/tasks.rs
Buckram123: Oops, just noticed that there is also "task_hash" string as well) Anyway do we want to format it in some different way?

contracts/cw-croncat/src/contract.rs
```diff
@@ -201,7 +201,6 @@ impl<'a> CwCroncat<'a> {
                 to_binary(&self.query_get_tasks_by_owner(deps, owner_id)?)
             }
             QueryMsg::GetTask { task_hash } => to_binary(&self.query_get_task(deps, task_hash)?),
-            QueryMsg::GetTaskHash { task } => to_binary(&self.query_get_task_hash(*task)?),
```
TrevorJTClarke, line 204: Hey @deveusss, im not happy with removing this. Do you see duplicated way of getting task hash? External integrations (web, etc) should be able to compute tash hash with query, so we have a reliable way without additional libs or self-compute that could be innacurate.
If this isn't duplicated, we need to keep this as the website uses it.

TrevorJTClarke (changes requested):

contracts/cw-croncat/src/contract.rs
Buckram123: @TrevorJTClarke I think it's still supposed to be removed in #209 right?

TrevorJTClarke: I dont think so - adding simulate is different use case thank just purely getting task_hash. 
Having tash_hash part of simulate is def a nice thing!

deveusss: brought GetTaskHash back

contracts/cw-croncat/src/tasks.rs
mikedotexe: Curious why we're not using `.add_attribute` here, @deveusss?

deveusss: @mikedotexe  image is not loading)

deveusss: @mikedotexe The reason is the consumer, like vectis, could get data from events

mikedotexe (approved):

TrevorJTClarke (approved):

contracts/cw-croncat/src/tasks.rs
TrevorJTClarke: @deveusss i still see format! in there, can you change to only hash? don't want vectis or others having to string split `task_hash jfdiksajfkldsjfdkslajf`

TrevorJTClarke (changes requested):

contracts/cw-croncat/src/tasks.rs
deveusss: thank you, done)

TrevorJTClarke (approved):

PR #217, opened 2022-12-15
Tick deletes only active agents

### LINKED ISSUE #215: [BUG] Tick Task removes all pending agents
### Goal:
Only assess active agents during the `tick` method.

### Problem:
Currently the logic iterates over all agents, regardless of active status. Effectively, running tick fully stops croncat runtime. 

### Requirements:
* Change this code: https://github.com/CronCats/cw-croncat/blob/main/contracts/cw-croncat/src/manager.rs#L526-L551
* Logic:
  * The loop should only iterate the active agents list
  * add a final check after the loop (after line 545) to check if:
    * A: No active agents, have pending agents, nominate first pending agent
    * B: No active && no pending, add emergency attribute for our indexer to pick up (see spec below)

### Tick empty spec
```
.add_attribute("lifecycle", "tick_failure")
```

NinoLipartiia:
Close #215

- `tick` doesn't remove pending agents
- added attribute if there aren't any active or pending agents
- I didn't add nomination. I think we already nominating the first pending agent when we call `unregister_agent` (checked it in test)


TrevorJTClarke (approved):
This looks perfect!! Thank you @NinoLipartiia!!

contracts/cw-croncat/src/tests/manager.rs
```diff
@@ -1928,14 +1997,43 @@ fn tick() {
         .attributes
         .iter()
         .any(|attr| attr.key == "account_id" && attr.value == AGENT0)));
+    assert!(!res.events.iter().any(|ev| ev
+        .attributes
+        .iter()
+        .any(|attr| attr.key == "lifecycle" && attr.value == "tick_failure")));
```
Buckram123, line 2003: Nice extra check in tests here as well!

Buckram123 (approved):
LGTM

PR #233, opened 2022-12-21
[FIX] Tasks with no actions can be created

### LINKED ISSUE #222: [BUG] Tasks with no actions can be created
Thought this was covered, but seems weird to allow empty actions array. Unless you really really wanna donate in a spammy way to agents. 😹

deveusss:
closes #222


contracts/cw-croncat/src/tasks.rs
```diff
@@ -254,7 +254,9 @@ impl<'a> CwCroncat<'a> {
                 val: "Interval invalid".to_string(),
             });
         }
-
+        if task.actions.is_empty() {
+            return Err(ContractError::CoreError(CoreError::InvalidAction {}));
+        }
```
Buckram123, line 259: What if we move it inside `is_valid_msg_calculate_usage()` task's method?

Buckram123 (changes requested):
Let's add a little test, just to make sure it's covered

Buckram123 (approved):
Nice to have that covered

PR #241, opened 2022-12-22
CronCat automatically executes passed DAO DAO proposals

### LINKED ISSUE #180: [FEATURE] Use case: CronCat automatically executes passed DAO DAO proposals
DAO DAO testnet is working again and is located here:
https://testnet.daodao.zone

For this task, please create a new DAO that only has one member: you.

I have been using the token-based DAO, but it's possible you can use the multisig one, too.

image

Go ahead and create a proposal for your testnet DAO and then vote "Yes" on it to pass it. After you do that, there will be an "Execute" button that appears in the DAO DAO site. Click on that, but don't sign it yet. Look at the "Data" tab in Keplr.

Screen Shot 2022-10-26 at 1 55 59 PM

When you click on the Execute button you see this:

Screen Shot 2022-10-26 at 1 57 00 PM

The `msg` field contains a Base64 value. You can use [Boop](https://github.com/IvanMathy/Boop) or another tool to decode it. You'll see it's a JSON string, which, when formatted, looks like this:

```json
{
  "execute": {
    "proposal_id": 3
  }
}
```

Anyone can execute a proposal once it's passed. We want a CronCat task to be the one that executes the proposal.

Also please note that the `contract` is different from the DAO address. DAO DAO uses many different smart contracts in a modular way, and this is one of the contracts dealing with proposals. As you are working on this ticket, remember that you can easily get that contract address from Keplr, as it takes a bit longer to find it otherwise.

---

In case it's helpful, here's a CLI command that will execute this proposal, as well.

`junod tx wasm execute juno1v5j9xglgjqsrs8z9mus7jawvafn3y4l2q52y3xu0ypdf3mm09c7qegqx9g '{"execute":{"proposal_id":3}}' --node https://juno-testnet-rpc.polkachu.com:443 --chain-id uni-5 -o json --from mikereg --gas-prices 0.025ujunox --gas auto --gas-adjustment 1.3 --broadcast-mode block | jq`

But we obviously want to write this as a CronCat task, as an Action.

You can perhaps create a script to help, that will be similar to this one, except we'd change out the Action here:
https://github.com/CronCats/cw-croncat/blob/deccdae184f720445875ae41e39e3461f32f10b6/scripts/uni-testnet/create_staking_task.sh#L29-L37

and instead we'd put the message shown above in JSON, and we'll add a rule to check that the proposal has passed.

As you recall, we have code checking if a proposal has passed here:
https://github.com/CronCats/cw-croncat/blob/deccdae184f720445875ae41e39e3461f32f10b6/contracts/cw-rules/src/contract.rs#L169-L180

While we're talking about using the `junod` tool, here's a command that will create a proposal for my testnet DAO. Feel free to use this and modify the values for the next section, when you need to quickly create proposals to test this:

`junod tx wasm execute juno1tm2eg6zzqy6mhf9d0khgdm6uegdxlzdfx09xydjpk0rvcm9agwnsel0lks '{"propose":{"msg":{"propose":{"title":"see it works","description":"from the command-line","msgs":[]}}}}' --node https://juno-testnet-rpc.polkachu.com:443 --chain-id uni-5 -o json --from mikereg --gas-prices 0.025ujunox --gas auto --gas-adjustment 1.3 --broadcast-mode block | jq`

---

Note: for this preliminary use case, I think it's fair to say that a task is created for each proposal. We will not (for now) assume that a DAO wants *every* proposal to be automatically executed. We will talk to the DAO DAO team more about this, but for now let's assume that once we show this is working for a single proposal, perhaps an end user can have a new option like this:

Screen-Shot-2022-10-26-at-2 12 11-PM

In Photoshop I created what I imagine will be a feature in DAO DAO, where it says "Check this to have proposal automatically executed if passed"

I wanted to make that to illustrate that this will be per-proposal for now, and until we hear otherwise. Let's build and test this feature before we have that checkbox, and have it work on your personal testnet DAO DAO. Since we don't have the checkbox, what that will look like is you creating a proposal, getting the proposal ID, and then creating a task similar to the bash script from above. Then you'll vote "Yes" on your proposal so its status changes to "passed" and then the task should execute because the "rule" has evaluated to true.

NinoLipartiia:
Closes #180


contracts/cw-croncat/src/helpers.rs
```diff
@@ -177,11 +177,7 @@ impl<'a> CwCroncat<'a> {
             // It's possible there are more "covered tasks" than total tasks,
             // so use saturating subtraction to hit zero and not go below
             let total_tasks_needing_agents = total_tasks.saturating_sub(num_tasks_covered);
-            let remainder = if total_tasks_needing_agents % max_tasks == 0 {
-                0
-            } else {
-                1
-            };
+            let remainder = u64::from(total_tasks_needing_agents % max_tasks != 0);
```
mikedotexe, line 180: 👍🏼

Buckram123: coding in C flashbacks 🫥

contracts/cw-rules/src/contract.rs
```diff
@@ -203,6 +208,36 @@ fn query_dao_proposal_status(
...
+    let dao_addr = deps.api.addr_validate(&dao_address)?;
+    // Query the amount of proposals
+    let proposal_count = deps
+        .querier
+        .query_wasm_smart(dao_addr.clone(), &QueryDao::ProposalCount {})?;
+    let res: ProposalListResponse = deps.querier.query_wasm_smart(
+        dao_addr,
+        &QueryDao::ListProposals {
+            start_after: None,
+            limit: Some(proposal_count),
```
Buckram123, line 223: Wonder at what point it would break. Should we make `start_after` optionally customizable?

contracts/cw-croncat/src/balancer.rs
```diff
@@ -292,3 +289,9 @@ impl<'a> Balancer<'a> for RoundRobinBalancer {
         Ok(())
     }
 }
+
+impl Default for RoundRobinBalancer {
+    fn default() -> RoundRobinBalancer {
+        RoundRobinBalancer::new(BalancerMode::ActivationOrder)
+    }
+}
```
Buckram123, line 297: Nice

Buckram123 (approved):
LGTM

contracts/cw-croncat/src/helpers.rs
NinoLipartiia: Agree, but that's what clippy suggested :)

mikedotexe:
Thank you!

PR #242, opened 2022-12-22
[FIX]: Optimiser bug "does not match the specified platform"

deveusss:


justfile
```diff
@@ -54,7 +54,6 @@ optimize-m1:
 	docker run --rm -v "$(pwd)":/code \
 		--mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
 		--mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
-		--platform linux/arm64 \
 		cosmwasm/workspace-optimizer:0.12.8
```
Buckram123, line 57: shouldn't it be ?
```
cosmwasm/workspace-optimizer-arm64
```

Buckram123 (changes requested):

justfile
deveusss: DANGER
There is an [rust-optimizer-arm64](https://hub.docker.com/r/cosmwasm/rust-optimizer-arm64) docker builder image that gives faster output on Apple M1 chip. Arm images are released to ease development and testing on Mac M1 machines. The native Arm version produces different wasm artifacts than the Intel version. For release / production use, only contracts built with the Intel optimizers must be used.

Buckram123: > DANGER There is an [rust-optimizer-arm64](https://hub.docker.com/r/cosmwasm/rust-optimizer-arm64) docker builder image that gives faster output on Apple M1 chip. Arm images are released to ease development and testing on Mac M1 machines. The native Arm version produces different wasm artifacts than the Intel version. For release / production use, only contracts built with the Intel optimizers must be used.

We're not planning to deploy m1 wasm files, it's only for testing, right? Or what's the purpose of `optimize-m1` command then?

deveusss: for testing localy

Buckram123: @deveusss but rust ones using arm64 image https://github.com/CronCats/cw-croncat/blob/7c7230d889a170a9c194c654689efb2eb0667e5e/justfile#L59-L69

Buckram123: and i'm still leaning towards removing the rust ones https://github.com/CronCats/cw-croncat/pull/224#discussion_r1054357410

deveusss: @Buckram123 what does it mean, rust optimiser can use **arm** ones

PR #248, opened 2023-01-05
Update mod-nft

NinoLipartiia:
Created a separate contract for nft rules.
Added `AddrHasNft`, which checks if the address has any tokens on this `nft_contract`.

Also according to [this documentation](https://github.com/CronCats/cw-croncat/wiki/%E2%9A%99%EF%B8%8F-Rules-Architecture#nft-rules) there should be two more queries, which check if `token_id` doesn't have owner. But i think `owner` [is not optional](https://github.com/CosmWasm/cw-nfts/blob/c7be7aba9fb270abefee5a3696be62f2736592a0/contracts/cw721-base/src/state.rs#L105) and assigned as soon as the token is minted. Could you please explain how these queries should work?


contracts/mod-nft/src/contract.rs
```diff
@@ -22,10 +33,75 @@ pub fn execute(
     _info: MessageInfo,
     _msg: ExecuteMsg,
 ) -> Result<Response, ContractError> {
-    todo!();
+    Err(ContractError::Noop)
```
Buckram123, line 36:
```suggestion
    Err(ModError::Noop)
```
Do we want to move that to the `mod-sdk` package? Every mod will have only this one error

contracts/mod-nft/src/error.rs
```diff
@@ -5,4 +5,7 @@ use thiserror::Error;
 pub enum ContractError {
     #[error("{0}")]
     Std(#[from] StdError),
```
Buckram123, line 7:
```suggestion
```
It can't return StdError, since we don't check anything

contracts/mod-nft/src/msg.rs
```diff
@@ -8,4 +11,13 @@ pub enum ExecuteMsg {}
...
-pub enum QueryMsg {}
+pub enum QueryMsg {
+    #[returns(QueryResponse)]
+    OwnerOfNft(OwnerOfNft),
+
+    #[returns(QueryResponse)]
+    AddrHasNft {
+        address: String,
+        nft_address: String,
+    },
```
Buckram123, line 22: Don't you mind adding short descriptions to the enum variants?

contracts/mod-nft/src/contract.rs
```diff
@@ -22,10 +33,75 @@ pub fn execute(
...
+        valid_nft,
+        &Tokens {
+            owner: address,
+            start_after: None,
+            limit: None,
+        },
+    )?;
+    Ok(QueryResponse {
+        result: !res.tokens.is_empty(),
+        data: Default::default(),
```
Buckram123, line 105: I think returning tokens might be useful here!
So we can avoid another query of list of nfts :)

contracts/mod-nft/src/contract.rs
```diff
@@ -22,10 +33,75 @@ pub fn execute(
...
+    address: String,
+    nft_address: String,
+) -> StdResult<QueryResponse> {
+    let valid_nft = deps.api.addr_validate(&nft_address)?;
+    let res: TokensResponse = deps.querier.query_wasm_smart(
+        valid_nft,
+        &Tokens {
+            owner: address,
+            start_after: None,
+            limit: None,
```
Buckram123, line 100: Thinking if we should add limit here. I know default limit should be covered by the cw721 itself, but what if it fails?

Note: if we consider to not return tokens, we want it to be exactly "1"

contracts/mod-nft/src/contract.rs
```diff
@@ -22,10 +33,75 @@ pub fn execute(
...
+    address: String,
+    nft_address: String,
+    token_id: String,
+) -> StdResult<QueryResponse> {
+    let valid_nft = deps.api.addr_validate(&nft_address)?;
+    let res: OwnerOfResponse = deps.querier.query_wasm_smart(
+        valid_nft,
+        &OwnerOf {
+            token_id,
+            include_expired: None,
```
Buckram123, line 73: Can this field be useful in any way?

Buckram123 (changes requested):
Nice progress! Small nits

contracts/mod-nft/Cargo.toml
```diff
@@ -17,6 +17,11 @@ library = []
...
 cosmwasm-std = { workspace = true }
 cosmwasm-schema = { workspace = true }
+cw2 = { workspace = true }
+cw721 = { workspace = true }
+cw-multi-test = { workspace = true }
+mod-sdk = { workspace = true }
+cw721-base = { workspace = true }
 thiserror = { workspace = true }
 
 [dev-dependencies]
```
Buckram123, line 27:
```suggestion
mod-sdk = { workspace = true }
thiserror = { workspace = true }

[dev-dependencies]
cw721-base = { workspace = true }
cw-multi-test = { workspace = true }
```
Missed last thing

contracts/mod-nft/src/contract.rs
NinoLipartiia: Do you think we should move `ContractError` to `mod-sdk` and include only the error about `execute`?

Buckram123: > Do you think we should move `ContractError` to `mod-sdk` and include only the error about `execute`?

I like the idea, but I'm not sure) We should call it something like `ModError` instead if we do so

contracts/mod-nft/src/error.rs
NinoLipartiia: But we still can get `StdError` inside queries, can't we?

contracts/mod-nft/src/contract.rs
```diff
@@ -1,18 +1,29 @@
-#[cfg(not(feature = "library"))]
-use cosmwasm_std::entry_point;
+use cosmwasm_std::{entry_point, to_binary, StdError};
```
Buckram123, line 1:
```suggestion
#[cfg(not(feature = "library)
use cosmwasm_std::entry_point;
use cosmwasm_std::{to_binary, StdError};
```
Just to be safe!

contracts/mod-nft/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 pub mod contract;
-pub mod error;
 pub mod msg;
-pub mod state;
+#[cfg(test)]
+mod tests;
+pub mod types;
```
Buckram123, line 5:
```suggestion
pub mod types;

pub use mod_sdk::error::ModError as ContractError;
```
Sorry for this :D Let's be consistent and call it as `ContractError` and make it public inside `lib.rs`

contracts/mod-nft/src/tests/has_nft.rs
```diff
@@ -0,0 +1,82 @@
...
+    let msg = QueryMsg::AddrHasNft {
+        address: ADMIN.to_string(),
+        nft_address: cw721_contract.to_string(),
+    };
+    let res: QueryResponse = app
+        .wrap()
+        .query_wasm_smart(contract_addr.clone(), &msg)
+        .unwrap();
+    assert!(!res.result);
+    assert_eq!(res.data, to_binary::<std::vec::Vec<String>>(&vec![])?);
```
Buckram123, line 71: Thanks for testing data!

Funny you had to do that cast

Buckram123 (approved):
Good job on cleaning up!

contracts/mod-nft/src/contract.rs
```diff
@@ -22,10 +36,75 @@ pub fn execute(
...
+            include_expired: None,
+        },
+    )?;
+    Ok(QueryResponse {
+        result: address == res.owner,
+        data: to_binary(&res.owner)?,
+    })
+}
+
+/// Query: OwnerOfNft
```
TrevorJTClarke, line 85: I think this naming might be confusing. Since the condition is `owns any`, can we change to be `OwnerHasNft` or `OwnerNfts`

contracts/mod-nft/src/contract.rs
```diff
@@ -22,10 +36,75 @@ pub fn execute(
...
+/// Used as a helper method to check if address owns any nft tokens
+///
+/// Response: QueryResponse
+/// Returns true if address owns at least one token for this contracts
+/// Data is empty
```
TrevorJTClarke, line 90: Could we optionally return the nft holdings? Could be rad for future transforms

TrevorJTClarke (approved):
Looks solid @NinoLipartiia!!
Minor nit pick on naming convention, otherwise GTG!! ✨

contracts/mod-nft/src/contract.rs
NinoLipartiia: @TrevorJTClarke  `OwnerOfNft` checks whether `address` is the owner of particular nft, so I think `OwnerHasNft` sounds more confusing.

`Owns any` is the condition for `AddrHasNft`, it checks if `address` has at least one nft

TrevorJTClarke: Its just a simple typo i think -- its a copy/paste from above

PR #249, opened 2023-01-05
Tests for mod-balances

NinoLipartiia:


contracts/mod-balances/Cargo.toml
```diff
@@ -19,7 +19,10 @@ cosmwasm-std = { workspace = true }
...
 cw2 = { workspace = true }
 cw20 = { workspace = true }
-thiserror = { workspace = true }
+cw20-base = { workspace = true }
+cw-multi-test = { workspace = true }
+cw-utils = { workspace = true }
 mod-sdk = { workspace = true }
+thiserror = { workspace = true }
 
 [dev-dependencies]
```
Buckram123, line 28:
```suggestion
mod-sdk = { workspace = true }
thiserror = { workspace = true }

[dev-dependencies]
cw20-base = { workspace = true }
cw-multi-test = { workspace = true }
cw-utils = { workspace = true }
```
We don't test dependencies to be included outside tests

contracts/mod-balances/src/msg.rs
```diff
@@ -13,15 +13,12 @@ pub enum ExecuteMsg {}
...
-        denom: String,
-    },
+    GetBalance { address: String, denom: String },
     #[returns(QueryResponse)]
     GetCw20Balance {
         cw20_contract: String,
         address: String,
     },
     #[returns(QueryResponse)]
     HasBalanceComparator(HasBalanceComparator),
```
Buckram123, line 23: Let's add small descriptions to enum variants!

contracts/mod-balances/src/tests/balance_comparator.rs
```diff
@@ -0,0 +1,353 @@
...
+            NATIVE_DENOM.to_string(),
+        ))),
+        comparator: BalanceComparator::Ne,
+    });
+    let res: QueryResponse = app
+        .wrap()
+        .query_wasm_smart(contract_addr.clone(), &msg)
+        .unwrap();
+    assert!(res.result);
+    assert_eq!(res.data, to_binary(&Uint128::new(1_000_000)).unwrap());
```
Buckram123, line 183: Thanks for testing `QueryResponse.data`!

Buckram123 (approved):
Good job, few fixes and LGTM

contracts/mod-balances/src/tests/balance_comparator.rs
```diff
@@ -0,0 +1,353 @@
...
+            NATIVE_DENOM.to_string(),
+        ))),
+        comparator: BalanceComparator::Lt,
+    });
+    let res: QueryResponse = app
+        .wrap()
+        .query_wasm_smart(contract_addr.clone(), &msg)
+        .unwrap();
+    assert!(!res.result);
+    assert_eq!(res.data, to_binary(&Uint128::new(1_000_000)).unwrap());
```
TrevorJTClarke, line 119: Need a passing test for `Lt` after this

contracts/mod-balances/src/tests/balance_comparator.rs
```diff
@@ -0,0 +1,353 @@
...
+    });
+    let res: QueryResponse = app
+        .wrap()
+        .query_wasm_smart(contract_addr.clone(), &msg)
+        .unwrap();
+    assert!(res.result);
+    assert_eq!(res.data, to_binary(&Uint128::from(15u128)).unwrap());
+
+    Ok(())
+}
```
TrevorJTClarke, line 353: Same as with Native, needs coverage of `Gt` & `Gte`

TrevorJTClarke (changes requested):
Hey @NinoLipartiia this is looking awesome!!
Left a couple comments on tests cov missing, minor things.

I think i made a big mistake on ergonomics as well.
The comparison is backwards, so for example the logic is currently:
`required_balance <= balance` 
which feels wrong, i believe it needs to be this for all of them:
`balance <= required_balance`
So when integrating, the user specifying Lte, expects the balance to fluctuate and the requirement to stay consistent.

Happy to chat on that if it doesn't make sense.

TrevorJTClarke (approved):
YEAHHHHHHHHHH

PR #250, opened 2023-01-06
croncat-manager split

Buckram123:
TODO(After `tasks` and `agents` split done):
- finish `proxy_call()`, `tick()` and `schema`


contracts/croncat-manager/src/balances.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// Contract methods
+
+pub fn execute_receive_cw20(
+    deps: DepsMut,
+    info: MessageInfo,
+    wrapper: Cw20ReceiveMsg,
+) -> Result<Response, ContractError> {
+    let msg: ReceiveMsg = from_binary(&wrapper.msg)?;
```
Buckram123, line 127: Noticed that everyone uses `ReceiveMsg` even if it's empty. I guess it's better to be sure why user sent tokens here. 
https://github.com/CosmWasm/cw-plus/blob/4da476f9e426fb87689b6f0c3398ff08a65248d1/contracts/cw4-stake/src/msg.rs#L42-L48

contracts/croncat-manager/src/balances.rs
```diff
@@ -0,0 +1,269 @@
...
+            let coin_addr = info.sender;
+
+            let cw20_verified = Cw20CoinVerified {
+                address: coin_addr,
+                amount: wrapper.amount,
+            };
+            let user_cw20_balance = add_user_cw20(deps.storage, &sender, &cw20_verified)?;
+            let available_cw20_balance = add_available_cw20(deps.storage, &cw20_verified)?;
+            Ok(Response::new()
+                .add_attribute("action", "receive_cw20")
```
Buckram123, line 143: all attributes in execute and instantiate methods 
```json
{"key": "method", "value": "method_name"}
```
replaced by
```json
{"key": "action", "value": "method_name"}
```

contracts/croncat-manager/src/contract.rs
```diff
@@ -1,31 +1,308 @@
...
+
+    // Update state
+    CONFIG.save(deps.storage, &config)?;
+    for coin in info.funds {
+        add_available_native(deps.storage, &coin)?;
+    }
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    Ok(Response::new()
+        // TODO?:.add_attribute("config", format!("{:?}, &config"))
```
Buckram123, line 96: We have too many config items, should we pack them in single attribute? What's your thoughts?

contracts/croncat-manager/src/msg.rs
```diff
@@ -1,11 +1,81 @@
...
+    Tick {},
+}
 
 #[cw_serde]
 #[derive(QueryResponses)]
-pub enum QueryMsg {}
+pub enum QueryMsg {
+    /// Gets current croncat config
+    #[returns(Config)]
+    Config {},
```
Buckram123, line 62: `GetQueryMethod`, renamed to the `QueryMethod`

Buckram123:
Should I add in here minimal factory contract, that will instantiate all of the contract and just store addresses of contracts?

packages/croncat-sdk-core/src/types.rs
```diff
@@ -0,0 +1,211 @@
...
+    pub agents_eject_threshold: Option<u64>,
+    pub balancer: Option<RoundRobinBalancer>,
+    // pub treasury_id: Option<String>,
+}
+
+#[cw_serde]
+pub struct BalancesResponse {
+    pub available_native_balance: Vec<Coin>,
+    pub available_cw20_balance: Vec<Cw20CoinVerified>,
+}
```
Buckram123, line 124: This struct had field `native_denom`, I believe it should be gathered from the config instead

contracts/croncat-manager/src/contract.rs
TrevorJTClarke: Honestly probably not. Can use a query for that later

contracts/croncat-manager/src/contract.rs
```diff
@@ -1,31 +1,297 @@
...
+            native_balances,
+            cw20_balances,
+            address,
+        } => execute_move_balances(deps, info, native_balances, cw20_balances, address),
+        ExecuteMsg::ProxyCall { task_hash: None } => execute_proxy_call(deps, env, info),
+        ExecuteMsg::ProxyCall {
+            task_hash: Some(task_hash),
+        } => execute_proxy_call_with_queries(deps, env, info, task_hash),
+        ExecuteMsg::Receive(msg) => execute_receive_cw20(deps, info, msg),
+        ExecuteMsg::WithdrawCw20WalletBalances { cw20_amounts } => {
```
TrevorJTClarke, line 126: Is this only for users or agents? Let's make it super obvious: `UserWithdraw` or `TaskBalanceWithdraw` or `AgentWithdraw`

contracts/croncat-manager/src/contract.rs
```diff
@@ -1,31 +1,297 @@
...
-    _info: MessageInfo,
-    _msg: ExecuteMsg,
+    info: MessageInfo,
 ) -> Result<Response, ContractError> {
-    todo!();
+    let config: Config = CONFIG.load(deps.storage)?;
+    check_ready_for_execution(&info, &config)?;
+
+    // TODO: query agent to check if ready
+    // TODO: execute task
```
TrevorJTClarke, line 142: Assuming TBD ;)

TrevorJTClarke (changes requested):
This is great progress!! left a few changes below :)

contracts/croncat-manager/src/balances.rs
```diff
@@ -0,0 +1,352 @@
...
+            })
+        })
+        .collect::<StdResult<_>>()?;
+
+    // update user and croncat manager balances
+    let mut updated_user_cw20 = Vec::with_capacity(withdraws.len());
+    let mut updated_available_cw20 = Vec::with_capacity(withdraws.len());
+    for cw20 in withdraws.iter() {
+        let new_user_bal = sub_user_cw20(deps.storage, &wallet, cw20)?;
+        let new_avail_bal = sub_available_cw20(deps.storage, cw20)?;
```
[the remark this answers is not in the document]
Buckram123: https://github.com/CronCats/cw-croncat/blob/31bef0f969ae236bce868ec2817f2e29e4080f2d/contracts/croncat-manager/src/balances.rs#L119

Don't worry it's covered! It will return overflow error!

contracts/croncat-manager/src/balances.rs
```diff
@@ -0,0 +1,352 @@
...
+}
+
+/// Execute: MoveBalances
+/// Used by owner of the contract to move balances from the manager to another address
+pub fn execute_move_balances(
+    deps: DepsMut,
+    info: MessageInfo,
+    native_balances: Vec<Coin>,
+    cw20_balances: Vec<Cw20Coin>,
+    address: String,
```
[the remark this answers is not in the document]
Buckram123: So the available balances used only for treasuries, right? This whole time I thought it was a combine of all balances we have, my bad! We don't have treasury_fee implemented yet tho. Should we rename it to the `TREASURY_NATIVE_BALANCE` and `TREASURY_CW20_BALANCE`?

contracts/croncat-manager/src/balances.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    if new_bal.is_zero() {
+        AVAILABLE_CW20_BALANCE.remove(storage, &cw20.address);
+    } else {
+        AVAILABLE_CW20_BALANCE.save(storage, &cw20.address, &new_bal)?;
+    }
+    Ok(new_bal)
+}
+
+// pub(crate) fn add_agent_native(
```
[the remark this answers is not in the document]
Buckram123: > remove if not-needed?

It will be needed

contracts/croncat-manager/src/balances.rs
```diff
@@ -0,0 +1,352 @@
...
+                amount: cw20.amount,
+            })?,
+            funds: vec![],
+        });
+    }
+
+    Ok(Response::new()
+        .add_attribute("action", "move_balances")
+        .add_message(bank_msg)
+        .add_messages(cw20_messages))
```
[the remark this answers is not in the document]
Buckram123: What do you mean by that @TrevorJTClarke

contracts/croncat-manager/src/state.rs
```diff
@@ -1 +1,34 @@
...
+/// Key: Cw20 Addr
+/// Value: Amount
+pub const AVAILABLE_CW20_BALANCE: Map<&Addr, Uint128> = Map::new("available_cw20_balance");
+
+// Accrued Agent reward balance
+pub const AGENT_BALANCES_NATIVE: Map<(&Addr, &str), Uint128> = Map::new("agent_balances_native");
+
+pub const AGENT_BALANCES_CW20: Map<(&Addr, &Addr), Uint128> = Map::new("agent_balances_cw20");
+
+pub const USERS_BALANCES_CW20: Map<(&Addr, &Addr), Uint128> = Map::new("users_balances_cw20");
```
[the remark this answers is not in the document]
Buckram123: Yes, we can't avoid storing it, unless we are okay to transfer balances from `tasks` contract

contracts/croncat-manager/src/balances.rs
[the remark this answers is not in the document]
TrevorJTClarke: That's a good thought. So to be 100% clear, we have 3 buckets of funds:
1. Agent rewards (base compensation for task fee, additional fee for profit)
2. Deposit funds (task funds to be used for compensation)
3. Owner/Treasury funds (balances earned to DAO similar to agent)

I am just hoping the implementation is extremely obvious with where allocation of funds are

packages/croncat-sdk-core/src/types.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub agent_fee: u64,
+    pub gas_price: GasPrice,
+    pub gas_base_fee: u64,
+    pub gas_action_fee: u64,
+    pub gas_query_fee: u64,
+    pub gas_wasm_query_fee: u64,
+    pub slot_granularity_time: u64,
+
+    // Treasury
+    // pub treasury_id: Option<Addr>,
```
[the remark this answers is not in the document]
Buckram123: We don't have it in our `main` branch, is there any requirements i can look at?

contracts/croncat-manager/src/balances.rs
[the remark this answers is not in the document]
TrevorJTClarke: so if response having empty vec messages, does it remove gas if we just start a mut response over the logic of this function? or am i crazy

packages/croncat-sdk-core/src/types.rs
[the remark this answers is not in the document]
TrevorJTClarke: the main has history from june when i did it and then pulled out :P 

Its pretty simple really:
IF treasury_id (or really `treasury_addr`) then any DAO fees/rewards can be withdrawn to the `treasury_addr`.
If NO `treasury_addr` on config, then withdraw goes to the `config.owner_id`

contracts/croncat-manager/src/state.rs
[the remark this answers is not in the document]
TrevorJTClarke: okay!! maybe just need to change name to be `temp` or similar

contracts/croncat-manager/src/balances.rs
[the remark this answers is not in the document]
Buckram123: There will be at least 1 message of `bank_msg`, if `cw20_messages` will be empty it will just do nothing

contracts/croncat-manager/src/helpers.rs
```diff
@@ -0,0 +1,19 @@
...
+use croncat_sdk_core::types::Config;
+
+use crate::ContractError;
+
+/// Check if contract is paused or user attached redundant funds.
+/// Called before every method, except [crate::contract::execute_update_config]
+pub(crate) fn check_ready_for_execution(
+    info: &MessageInfo,
+    config: &Config,
+) -> Result<(), ContractError> {
```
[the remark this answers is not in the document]
Buckram123: > Would this be better in packages?

I thought that we try to keep all contracts logic inside contract crate, unless it's a type logic?

packages/croncat-sdk-core/src/types.rs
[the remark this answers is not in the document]
Buckram123: @TrevorJTClarke what's our treasury_fees then?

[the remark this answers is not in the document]
TrevorJTClarke: @Buckram123 we will need to have same setup to agent fee - for now create a separate configurable variable, but it can be set to same value as agent

contracts/croncat-manager/src/helpers.rs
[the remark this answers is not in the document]
TrevorJTClarke: Interesting, good point. So its interesting, we could actually have some package for helpers/utils similar to: https://github.com/DA0-DA0/dao-contracts/blob/main/packages/cw-paginate/src/lib.rs

contracts/croncat-manager/src/contract.rs
```diff
@@ -113,23 +113,35 @@ pub fn execute(
...
         ExecuteMsg::ProxyCall { task_hash: None } => execute_proxy_call(deps, env, info),
         ExecuteMsg::ProxyCall {
             task_hash: Some(task_hash),
         } => execute_proxy_call_with_queries(deps, env, info, task_hash),
         ExecuteMsg::Receive(msg) => execute_receive_cw20(deps, info, msg),
-        ExecuteMsg::WithdrawCw20WalletBalances { cw20_amounts } => {
-            execute_withdraw_wallet_balances(deps, info, cw20_amounts)
-        }
+        ExecuteMsg::RefillNativeBalance {} => execute_refill_native_balance(deps, info),
+        ExecuteMsg::UserWithdraw {
```
TrevorJTClarke, line 126: TY!

packages/croncat-sdk-core/src/msg.rs
```diff
@@ -39,17 +39,21 @@ pub enum ManagerExecuteMsg {
     /// Note: it's shared across contracts
     UpdateConfig(UpdateConfig),
     /// Move balances from the manager to the address
-    MoveBalances {
+    OwnerWithdraw {
```
TrevorJTClarke, line 42: I still think it should not accept args at all, withdraw the whole amount

TrevorJTClarke (changes requested):

contracts/croncat-manager/src/contract.rs
```diff
@@ -113,23 +113,35 @@ pub fn execute(
...
+            native_balances,
+            cw20_balances,
+        } => execute_user_withdraw(deps, info, native_balances, cw20_balances),
         ExecuteMsg::Tick {} => execute_tick(deps, env, info),
     }
 }
 
+fn execute_refill_native_balance(
```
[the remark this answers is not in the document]
Buckram123: @TrevorJTClarke oops, need to take a better name for refilling `temp` users, any ideas?

packages/croncat-sdk-core/src/msg.rs
Buckram123: > I still think it should not accept args at all, withdraw the whole amount

Just to be clear, this method for withdrawing treasuries only, right? So it only should withdraw single native coin?

TrevorJTClarke: It should get called by owner (the DAO) with no args, and withdraw ALL funds available, both native, cw20, etc

Buckram123: > It should get called by owner (the DAO) with no args, and withdraw ALL funds available, both native, cw20, etc

By "available" what do you mean? Tasks balances/agents rewards gets stolen too?

contracts/croncat-manager/src/contract.rs
```diff
@@ -1,31 +1,297 @@
...
+        .add_attribute("paused", new_config.paused.to_string())
+        .add_attribute("owner_id", new_config.owner_id.to_string()))
+}
+
+/// Execute: UpdateConfig
+/// Helps manage and cleanup agents
+/// Deletes agents which missed more than agents_eject_threshold slot
+///
+/// Returns removed agents
+// TODO: It might be not possible to deserialize all of the active agents, need to find better solution
```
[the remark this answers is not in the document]
mikedotexe: We talked about a few options in today's meeting, but felt that we could set some sane limit for our early versions.

How's a limit of 5,000 agents sound?

contracts/croncat-manager/src/helpers.rs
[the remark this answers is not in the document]
mikedotexe: This is a pretty small function, and since it needs to load the config, feels like it's fine to have here imo

contracts/croncat-manager/src/contract.rs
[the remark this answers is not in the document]
Buckram123: > How's a limit of 5,000 agents sound?

How do we predict gas costs with such big limit? Should we add some gas_fee for each active_agent?

contracts/croncat-manager/src/helpers.rs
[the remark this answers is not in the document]
Buckram123: @mikedotexe I guess it could make sense to make a shared trait ConfigChecker or something like that, but yeah, it's ok to keep it here, at least for now imo :)

deveusss:
Very cool update!

contracts/croncat-manager/src/balances.rs
```diff
@@ -184,60 +64,111 @@ pub fn execute_receive_cw20(
...
             let user_cw20_balance = add_user_cw20(deps.storage, &sender, &cw20_verified)?;
-            let available_cw20_balance = add_available_cw20(deps.storage, &cw20_verified)?;
             Ok(Response::new()
                 .add_attribute("action", "receive_cw20")
                 .add_attribute("cw20_received", cw20_verified.to_string())
-                .add_attribute("available_cw20_balance", available_cw20_balance)
                 .add_attribute("user_cw20_balance", user_cw20_balance))
         }
+        ReceiveMsg::RefillTaskBalance { task_hash } => {
+            // TODO: check if sender is task_owner
```
TrevorJTClarke, line 84: Def need to finish this TODO before I feel good merging

TrevorJTClarke: Is this waiting on the tasks contract impl? If so, it shouldn't change msg right, so we can utilize the previous msg here?

contracts/croncat-manager/src/balances.rs
```diff
@@ -184,60 +64,111 @@ pub fn execute_receive_cw20(
...
+pub fn execute_refill_task_cw20(
+    deps: DepsMut,
+    info: MessageInfo,
+    task_hash: String,
+    cw20: Cw20Coin,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    check_ready_for_execution(&info, &config)?;
+
+    // TODO: check if sender is task_owner
```
TrevorJTClarke, line 117: here too please!

contracts/croncat-manager/src/balances.rs
```diff
@@ -255,109 +186,69 @@ pub fn execute_user_withdraw(
...
-        .into_iter()
-        .map(|balance| {
-            let coin = Cw20CoinVerified {
-                address: deps.api.addr_validate(&balance.address)?,
-                amount: balance.amount,
-            };
-            Ok(coin)
-        })
-        .collect::<StdResult<Vec<Cw20CoinVerified>>>()?;
+    let address = config.treasury_addr.unwrap_or(config.owner_addr);
```
TrevorJTClarke, line 201: thank you!

contracts/croncat-manager/src/balances.rs
```diff
@@ -255,109 +186,69 @@ pub fn execute_user_withdraw(
...
-        .skip(from_index as usize)
-        .take(limit as usize)
-        .map(|balance_res| {
-            balance_res.map(|(address, amount)| Cw20CoinVerified { address, amount })
-        })
-        .collect::<StdResult<Vec<Cw20CoinVerified>>>()?;
+    if config.paused {
+        return Err(ContractError::Paused {});
+    }
+    // TODO: query tasks to check it's task owner who refills
```
TrevorJTClarke, line 228: here please :)

contracts/croncat-manager/src/balances.rs
```diff
@@ -255,109 +186,69 @@ pub fn execute_user_withdraw(
...
-    })
+    if info.funds.len() > 2 {
+        return Err(ContractError::TooManyCoins {});
+    }
+    for coin in info.funds {
+        if coin.denom == config.native_denom {
+            task_balances.native_balance += coin.amount
+        } else {
+            let mut ibc = task_balances
+                .ibc_coin
```
TrevorJTClarke, line 241: is this new?

contracts/croncat-manager/src/contract.rs
```diff
@@ -62,14 +51,14 @@ pub fn instantiate(
         return Err(ContractError::InvalidGasPrice {});
     }
 
-    let owner_id = owner_id
+    let owner_addr = owner_addr
```
TrevorJTClarke, line 54: this is a great idea

contracts/croncat-manager/src/error.rs
```diff
@@ -20,4 +20,10 @@ pub enum ContractError {
 
     #[error("Must not attach funds")]
     RedundantFunds {},
+
+    #[error("Only up to one ibc coin supported")]
```
TrevorJTClarke, line 24: this seems odd - maybe you could help me understand why?

packages/croncat-sdk-manager/src/msg.rs
```diff
@@ -14,15 +13,7 @@ pub struct ManagerInstantiateMsg {
     /// Name of the key for raw querying Agents address from the factory
     pub croncat_agents_key: (String, [u8; 2]),
     /// Address of the contract owner, defaults to the sender
-    pub owner_id: Option<String>,
-    /// Gas needed to cover [ExecuteMsg::ProxyCall] without any action
-    pub gas_base_fee: Option<Uint64>,
-    /// Gas needed to cover single non-wasm task's Action
-    pub gas_action_fee: Option<Uint64>,
-    /// Gas needed to cover single non-wasm query
-    pub gas_query_fee: Option<Uint64>,
```
TrevorJTClarke: Can you explain why these were removed @Buckram123 ? Do they exist somewhere else? I think we still need this here

TrevorJTClarke (changes requested):
Hey @Buckram123 lookin great! Thank you for the latest changes, esp for treasury support.

One bigger question: How are we handling things if a user/contract simply sends (bank send) to the manager contract? would treasury be able to withdraw, or is it essentially locked?

contracts/croncat-manager/src/balances.rs
Buckram123: @TrevorJTClarke is it ok to finish it after merge? Can't test it yet)

contracts/croncat-manager/src/balances.rs
Buckram123: @TrevorJTClarke is it ok to finish it after merge? Can't test it yet)

contracts/croncat-manager/src/balances.rs
Buckram123: @TrevorJTClarke Yep, so we can have up to 1 ibc, up to 1 cw20 and native coins will be used in any case

packages/croncat-sdk-manager/src/msg.rs
Buckram123: > Can you explain why these were removed @Buckram123 ? Do they exist somewhere else? I think we still need this here

My vision here is to make manager keep economics, and gas should be handled by the tasks contract. So after task creation task passes message to the manager with gas to check if enough coins to the task, based of `gas_price`, `agent_fee` and `treasury_fee` configs

contracts/croncat-manager/src/error.rs
Buckram123: > this seems odd - maybe you could help me understand why?

Didn't we agree on up to 1 ibc and 1 cw20 coin per task?

contracts/croncat-manager/src/balances.rs
```diff
@@ -0,0 +1,281 @@
...
+    TASKS_BALANCES.save(deps.storage, task_hash.as_bytes(), &task_balances)?;
+    Ok(Response::new().add_attribute("action", "refill_native_balance"))
+}
+
+/// Query: Cw20WalletBalances
+/// Used to get user's available cw20 coins balance that he can use to attach to the task balance
+/// Can be paginated
+///
+/// Returns list of cw20 balances
+pub fn query_users_balances(
```
mikedotexe, line 259: I think I'd prefer if there were `cw20` somewhere in the method name, but not a massive deal

Buckram123: Agree, thanks

contracts/croncat-manager/src/contract.rs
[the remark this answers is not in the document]
mikedotexe: I don't think we're at the point of getting too concerned, we just need a limit

PR #251, opened 2023-01-10
Create mod-generic for generic queries

NinoLipartiia:


packages/mod-sdk/src/helpers.rs
```diff
@@ -0,0 +1,38 @@
+use cosmwasm_std::{to_vec, Binary, Deps, Empty, QueryRequest, StdError, StdResult, WasmQuery};
+use serde_cw_value::Value;
+
+pub fn query_wasm_smart_raw(
```
Buckram123, line 4: Let's move it to the `mod-generic`! I think we decided to keep the logic in the contracts, instead of packages?

Buckram123 (approved):
Nice one

packages/mod-sdk/src/helpers.rs
NinoLipartiia: I think we use `query_wasm_smart_raw` in mod-dao too, so I don't want to have it duplicated, what do you think?

NinoLipartiia: I changed `mod-dao` a bit, so it doesn't use `query_wasm_smart_raw` now. Moving `query_wasm_smart_raw` to `mod-generic`!

contracts/mod-generic/src/tests/generic_query.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use crate::tests::helpers::proper_instantiate;
+use crate::types::{GenericQuery, ValueIndex};
+use crate::value_ordering::ValueOrdering;
+
+#[test]
+fn test_generic() {
+    let (app, contract_addr, cw4_addr, _) = proper_instantiate();
+
+    let generic_query = GenericQuery {
```
TrevorJTClarke, line 15: Would love to see a comment here about how this is checking the weight of the second member is greater than 1.

contracts/mod-generic/src/tests/generic_query.rs
```diff
@@ -0,0 +1,85 @@
...
+            ValueIndex::Key("weight".to_string()),
+        ]
+        .into(),
+        ordering: ValueOrdering::UnitAbove,
+        value: to_binary(&1).unwrap(),
+        contract_addr: cw4_addr.into_string(),
+    };
+    let msg = QueryMsg::GenericQuery(generic_query);
+    let res: QueryResponse = app.wrap().query_wasm_smart(contract_addr, &msg).unwrap();
+    assert!(res.result);
```
TrevorJTClarke, line 33: Can we add a fail case too?

contracts/mod-generic/src/tests/generic_query.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+    });
+    let msg = generic_query_json.to_string().into_bytes();
+    let request = WasmQuery::Smart {
+        contract_addr: contract_addr.into(),
+        msg: Binary(msg),
+    }
+    .into();
+    let res: QueryResponse = app.wrap().query(&request).unwrap();
+    assert!(res.result);
```
TrevorJTClarke, line 61: Can we check the res.data as well?

contracts/mod-generic/src/tests/generic_query.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#[test]
+fn test_generic_bigint() {
+    let (app, contract_addr, _, cw20_addr) = proper_instantiate();
+
+    let generic_query = GenericQuery {
+        msg: to_binary(&cw20::Cw20QueryMsg::TokenInfo {}).unwrap(),
+        path_to_value: vec![ValueIndex::Key("total_supply".to_string())].into(),
+        ordering: ValueOrdering::UnitAbove,
```
TrevorJTClarke, line 71: Can we change to UnitBelow? Since we have UnitAbove in the tests above

TrevorJTClarke (changes requested):
Super rad stuff - only needs a couple more adjustments to tests!

PR #252, opened 2023-01-10
Update deps to dao v2.0.0 tag!

Buckram123:
Dao v2.0.0 just released! Let's have dependencies from the `tag` now :)


TrevorJTClarke (approved):
TYSM!

PR #253, opened 2023-01-11
Update mod-dao

NinoLipartiia:
Moved dao rules to the separate crate. 

Also added a new rule which checks for passed dao proposals with `Migrate` message (see [docs](https://github.com/CronCats/cw-croncat/wiki/%E2%9A%99%EF%B8%8F-Rules-Architecture#dao-rules)).

~~Also we don't have a rule that checks if current proposal id is greater than specified value, do we still need it? Instead there is a rule that checks for passed proposals and returns a list of them.~~ Added `HasNewProposals` message which checks if there are any proposals with index greater that `value`.


contracts/mod-dao/src/contract.rs
```diff
@@ -22,10 +35,181 @@ pub fn execute(
...
+        QueryMsg::CheckWithMigration { dao_address } => {
+            to_binary(&query_proposals_with_migration(deps, dao_address)?)
+        }
+        QueryMsg::HasNewProposals { dao_address, value } => {
+            to_binary(&query_has_new(deps, dao_address, value)?)
+        }
+    }
+}
+
+/// Query: CheckProposalStatus
```
TrevorJTClarke, line 66: Let's change this to be `MatchProposalStatus` or `ProposalStatusMatches`

contracts/mod-dao/src/contract.rs
```diff
@@ -22,10 +35,181 @@ pub fn execute(
...
+    let resp: ProposalResponse = deps
+        .querier
+        .query_wasm_smart(dao_addr, &QueryDao::Proposal { proposal_id })?;
+    Ok(QueryResponse {
+        result: resp.proposal.status == status,
+        data: to_binary(&resp)?,
+    })
+}
+
+/// Query: CheckPassedProposals
```
TrevorJTClarke, line 88: Let's change this to `HasPassedProposals`

contracts/mod-dao/src/contract.rs
```diff
@@ -22,10 +35,181 @@ pub fn execute(
...
+            });
+        }
+    }
+    Ok(QueryResponse {
+        result: false,
+        data: to_binary(&res.proposals)?,
+    })
+}
+
+/// Query: CheckWithMigration
```
TrevorJTClarke, line 122: Can we change to `HasPassedProposalWithMigration` or something?

TrevorJTClarke: How does this work in terms of gas?
How does this work in terms of if both `HasPassedProposals` and this method are in tandem? is it a race to execute?

contracts/mod-dao/src/contract.rs
```diff
@@ -22,10 +35,181 @@ pub fn execute(
...
+        }
+    }
+
+    Ok(QueryResponse {
+        result: !with_migration.is_empty(),
+        data: to_binary(&with_migration)?,
+    })
+}
+
+/// Query: HasNewProposals
```
TrevorJTClarke, line 198: Name is confusing: `HasNewProposals` sounds similar to the above, can we change to something similar: `HasProposalsGtId`

contracts/mod-dao/src/tests/passed_proposals.rs
```diff
@@ -0,0 +1,406 @@
...
+    let res: QueryResponse<Binary> = app
+        .wrap()
+        .query_wasm_smart(
+            contract_addr.clone(),
+            &QueryMsg::CheckPassedProposals {
+                dao_address: govmod_single.to_string(),
+            },
+        )
+        .unwrap();
+    assert!(!res.result,);
```
TrevorJTClarke, line 206: See a weird comment in the assert here

contracts/mod-dao/src/tests/passed_proposals.rs
```diff
@@ -0,0 +1,406 @@
...
+    // Query passed proposals and execute them
+    for num in 1..51 {
+        let index: u64 = 2 * num;
+
+        // Check that CheckPassedProposals returns index as the number of the first passed proposal
+        let res: QueryResponse<Binary> = app
+            .wrap()
+            .query_wasm_smart(
+                contract_addr.clone(),
+                &QueryMsg::CheckPassedProposals {
```
TrevorJTClarke, line 370: This is dope

contracts/mod-dao/src/tests/proposal_status.rs
```diff
@@ -0,0 +1,451 @@
...
+    // It's executed now
+    // Test if other types of status works
+    let res: QueryResponse = app
+        .wrap()
+        .query_wasm_smart(
+            contract_addr,
+            &QueryMsg::CheckProposalStatus(CheckProposalStatus {
+                dao_address: govmod_single.to_string(),
+                proposal_id: 1,
+                status: Status::Executed,
```
TrevorJTClarke, line 434: LOVE this too, have a few ideas this can be used for!! Like shipping NFTs or whatever

TrevorJTClarke (changes requested):
This is serious amazing! I am soooooo stoked for these features!!!
Left a couple naming convention things, just to help those be less confusing.

One nit pick, maybe we do later, could utilize more helpers in the tests.

Thank you @NinoLipartiia!!!

PR #254, opened 2023-01-12
Fix issue, when block slot starts at end of boundary

### LINKED ISSUE #244: [BUG] Block slot starts at end of boundary
Appears that block slots suffer from timestamp only granularity:

Screenshot 2022-12-22 at 3 49 33 PM

I was able to see this using:
```
{
  "get_slot_ids": {}
}
```

My question is:
Do we have a bug where if i schedule a task to run every 10 blocks, it trims the slot granularity via modulo such that it would wait for the next 100+ blocks later? i put a start block of `1355555` but the task never happened until block `1355650` then happened in rapid succession.

The rapid succession piece is confusing as well.

Here's the way to reproduce:
* 3 bank send actions
* Interval: Block(10)
* Start & end block, such that 10 occurrences

deveusss:
closes #244


mikedotexe:
Contracts deployed with this code:

### Rules
`juno1s8fv7dr3ppq2z4tlpqztt3ws3sljkqcp32yddknu0v98hs0rgyvqdqnu4j`

### Manager
`juno1pva373xs9grctllmmymp2jn0utxqlc25s049s6fh9evtggpgsdsqwv5ysp`

Testing now… and hopefully this will help others who want to play with it "live"

contracts/cw-croncat/src/tests/slots.rs
```diff
@@ -246,7 +246,7 @@ fn interval_get_next_block_by_offset() {
         ),
         (
             Interval::Block(100),
-            BoundaryValidated {
+            CheckedBoundary {
```
TrevorJTClarke, line 249: @deveusss do we have a test that covers both start/end? I only see start OR end

contracts/cw-croncat/src/tests/slots.rs
```diff
@@ -377,7 +377,7 @@ fn interval_get_next_cron_time() {
         // the task has ended
         (
             Interval::Cron("* * * * * *".to_string()),
-            BoundaryValidated {
+            CheckedBoundary {
                 start: None,
                 end: Some(1_571_797_419_879_305_532),
                 is_block_boundary: Some(false),
```
TrevorJTClarke, line 383: Same as above - i wanna see start & end combined in a test

TrevorJTClarke (approved):
BOOOOYAHHHHHH @deveusss 🤘

Small comment on adding tests, if you hazz extra min

contracts/cw-croncat/src/tests/slots.rs
deveusss: good point, will add missing one

mikedotexe:
Let this run 4 times under careful observation for a task that runs every minute. Working as expected now. Thanks!

Screen Shot 2023-01-12 at 4 20 42 PM (3)

mikedotexe (approved):

PR #259, opened 2023-01-13
Croncat factory

Buckram123:
Implementation based of https://github.com/CronCats/cw-croncat/wiki/%F0%9F%8F%AD-Factory-Architecture


contracts/croncat-factory/src/contract.rs
```diff
@@ -0,0 +1,370 @@
...
+fn query_versions_by_contract_name(
+    deps: Deps,
+    contract_name: String,
+) -> StdResult<Vec<ContractMetadataResponse>> {
+    let metadatas: Vec<(Vec<u8>, ContractMetadata)> = CONTRACT_METADATAS
+        .prefix(&contract_name)
+        .range(deps.storage, None, None, Order::Ascending)
+        .collect::<StdResult<_>>()?;
+
+    let mut versions = Vec::with_capacity(metadatas.len());
```
TrevorJTClarke, line 280: Would this ever require pagination? Kinda think no, but want to make sure theres no query gas limitations

contracts/croncat-factory/src/contract.rs
```diff
@@ -0,0 +1,370 @@
...
+            checksum: metadata.checksum,
+            changelog_url: metadata.changelog_url,
+            schema: metadata.schema,
+        };
+        versions.push(metadata_response);
+    }
+    Ok(versions)
+}
+
+pub fn query_latest_contracts(deps: Deps) -> StdResult<Vec<EntryResponse>> {
```
TrevorJTClarke, line 298: cant wait for this!!!

packages/croncat-sdk-factory/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "croncat-sdk-factory"
+version = "0.1.0"
+authors = []
```
TrevorJTClarke, line 4: feel free to add yourself

packages/croncat-sdk-factory/src/msg.rs
```diff
@@ -0,0 +1,139 @@
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use cosmwasm_std::{Addr, Binary};
+
+#[cw_serde]
+pub struct FactoryInstantiateMsg {
+    pub owner_addr: Option<String>,
+    pub manager_module_instantiate_info: ModuleInstantiateInfo,
```
TrevorJTClarke, line 7: still have reservations about this living here - i believe it should be specified during deploy only

packages/croncat-sdk-factory/src/msg.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub commit_id: String,
+
+    /// proof of deployed code
+    pub checksum: String,
+
+    /// public link to a README about this version
+    pub changelog_url: Option<String>,
+
+    /// types/schema - helps keep UI/clients backward compatible
+    pub schema: String,
```
TrevorJTClarke, line 90: Make this optional please :)

packages/croncat-sdk-factory/src/msg.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// Instantiate message to be used to create the contract.
+    pub msg: Binary,
+
+    /// Label for the instantiated contract.
+    pub label: String,
```
TrevorJTClarke, line 138: did you mean to change this to `contract_name`?

packages/croncat-sdk-factory/src/msg.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+#[cw_serde]
+pub struct EntryResponse {
+    pub contract_name: String,
+    pub metadata: ContractMetadataResponse,
+}
+
+#[cw_serde]
+pub struct ContractMetadata {
```
TrevorJTClarke, line 72: This is the same (basically) as `ModuleInstantiateInfo ` - the new ModuleInstantiateInfo has what we discussed on the call. Can you migrate to that for all instead?

TrevorJTClarke:
> * migrate with a little demo

Will definitely need a README on this contract :)

contracts/croncat-factory/src/contract.rs
```diff
@@ -0,0 +1,370 @@
...
+    Ok(Response::new().add_attribute("action", "update_metadata_changelog"))
+}
+
+fn execute_remove(
+    deps: DepsMut,
+    contract_name: String,
+    version: [u8; 2],
+) -> Result<Response, ContractError> {
+    // Can't remove latest
+    let latest_version = LATEST_VERSIONS
```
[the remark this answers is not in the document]
Buckram123: > Are you thinking this would be a future update?

Yep, thanks for noticing, i'd better add comment here.
Still not sure about the flow of checking that, obvious for manager contract, but what for others?

contracts/croncat-factory/src/contract.rs
```diff
@@ -0,0 +1,370 @@
...
+        QueryMsg::AllEntries {} => to_binary(&query_all_entries(deps)?),
+    }
+}
+
+fn query_all_entries(deps: Deps) -> StdResult<Vec<EntryResponse>> {
+    let metadatas: Vec<((String, Vec<u8>), ContractMetadata)> = CONTRACT_METADATAS
+        .range(deps.storage, None, None, Order::Ascending)
+        .collect::<StdResult<_>>()?;
+
+    let mut entries = Vec::with_capacity(metadatas.len());
```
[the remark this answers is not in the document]
Buckram123: Was thinking about it too! Depends on how much versions we plan to maintain.
I guess it could grow fast for library contracts

contracts/croncat-factory/src/state.rs
```diff
@@ -0,0 +1,27 @@
...
+pub const CONFIG: Item<Config> = Item::new("config");
+
+/// Contract name with the version to the metadata
+pub const CONTRACT_METADATAS: Map<(&str, &[u8]), ContractMetadata> = Map::new("contract_metadatas");
+
+/// Contract name with the version to the Addr
+pub const CONTRACT_ADDRS: Map<(&str, &[u8]), Addr> = Map::new("contract_addrs");
+
+/// Latest contract name to the Addr
+pub const LATEST_ADDRS: Map<&str, Addr> = Map::new("latest_addrs");
```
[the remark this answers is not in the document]
Buckram123: > Only question here is - is this more efficient than doing indexed map?

Most of the time spent here deciding the state :D Had so many options during it. And it comes to the preferences, really. 
Two things pushed me away from the indexed map: 
- Didn't want to store unchecked address in the state
- Want to be able to [safely rawquery](https://docs.rs/cw-storage-plus/latest/cw_storage_plus/struct.Map.html#method.query) it

packages/croncat-sdk-factory/src/msg.rs
Buckram123: > did you mean to change this to `contract_name`?

Like that name too! Not sure, how to not make it confusing. We can have both, one for chain(label) and other for the state(contract_name)

contracts/croncat-factory/src/contract.rs
```diff
@@ -0,0 +1,370 @@
...
+        code_id: init_info.code_id,
+        msg: init_info.msg,
+        funds: vec![],
+        label: init_info.label,
+    };
+    Ok(msg)
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn instantiate(
```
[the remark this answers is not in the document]
Buckram123: >What are the implications of this vs, new factory adopting old versions of previously deployed contracts? 

So you think just taking addresses of already deployed contracts in instantiate message will be fine? I guess having it 1 tx should be faster for re-deployments. Also it makes sure none of our instantiates of core contracts fails and we have factory that contains manager, agents, but not the tasks! Thoughts?

> Im inclined to simplify this instantiate so we cut down on complexity. I'm 100% okay with multiple TXNs for each new contract.

It sounds a little more complex to me if we take addr and it's more error prone instead of a freshly instantiated contract. I'm up for something like
```rust
pub struct InstantiateMsg{
 pub owner_addr: String,
 pub manager_module_instantiate_info: ModuleInfo,
 pub tasks_module_instantiate_info: ModuleInfo,
 pub agents_module_instantiate_info: ModuleInfo,
}

pub enum ModuleInfo{
 /// Needs to be instantiated
 ModuleInstantiateInfo {
 ...
 },
 /// Already instantiated contract
 ModuleAddr {
 contract_addr: String
 },
}
```
To cover both cases!

> Thoughts? If you agree, lets simplify instantiate down to only the base factory needs.

What's a minimal requirments for a factory needed? Can we roll with zero croncat contracts in the state?

contracts/croncat-factory/src/contract.rs
```diff
@@ -0,0 +1,370 @@
...
+    match msg {
+        ExecuteMsg::Deploy {
+            kind,
+            module_instantiate_info,
+        } => execute_deploy(deps, env, kind, module_instantiate_info),
+        ExecuteMsg::Remove {
+            contract_name,
+            version,
+        } => execute_remove(deps, contract_name, version),
+        ExecuteMsg::UpdateMetadataChangelog {
```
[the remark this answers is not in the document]
Buckram123: > What do you think about moving into just `UpdateMetadata`, and for now only updates changelog, so future integrations (if any) can remain simpler

I like that!

contracts/croncat-factory/src/contract.rs
```diff
@@ -0,0 +1,370 @@
...
+    let wasm = init_save_metadata_generate_wasm_msg(
+        deps.storage,
+        module_instantiate_info,
+        kind,
+        env.contract.address.as_str(),
+    )?;
+    let msg = SubMsg::reply_on_success(wasm, 0);
+
+    Ok(Response::new()
+        .add_attribute("action", "deploy")
```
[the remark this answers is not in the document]
Buckram123: > Thinking it would be nice to add attributes for version kind & contract_name

Thanks

contracts/croncat-factory/src/contract.rs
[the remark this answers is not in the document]
Buckram123: So I'll just add it, so we don't redeploy it, we are not losing anything anyway :)

contracts/croncat-factory/src/contract.rs
Buckram123: > Would this ever require pagination? Kinda think no, but want to make sure theres no query gas limitations

Unless croncat dao decides to break it - I don't think so. What's your thoughts on `Order`? Should we give latest versions first?

packages/croncat-sdk-factory/src/msg.rs
Buckram123: Right, only difference is `VersionKind`. I'm fine to specify it if we decide to deploy contracts one by one.
Small note if we decide to not instantiate all core contracts in a single tx: make paused default to the `true`

contracts/croncat-factory/src/contract.rs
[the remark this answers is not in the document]
TrevorJTClarke: Sorry, i think i miscommunicated a little.

The idea is that there is no logic for any of the modules/contracts in the factory instantiate. Only `pub owner_addr: String` basically

contracts/croncat-factory/src/contract.rs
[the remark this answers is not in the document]
TrevorJTClarke: I think it applies to any stateful contracts, since they will likely have a `paused` boolean in its state.

contracts/croncat-factory/src/contract.rs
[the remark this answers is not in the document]
TrevorJTClarke: okay thanks!

contracts/croncat-factory/src/contract.rs
TrevorJTClarke: ohhh good thought. Yeah definitely

contracts/croncat-factory/src/state.rs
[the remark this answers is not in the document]
TrevorJTClarke: thats right, yes i want rawquery to be priority on this!

packages/croncat-sdk-factory/src/msg.rs
TrevorJTClarke: I dont get the difference tho? Let's unify to 1 only, i think `contract_name` fits all cases

packages/croncat-sdk-factory/src/msg.rs
TrevorJTClarke: Thats fair. I would like to go non-instantiate route, all deploys use individual methods, outside of factory instantiate to keep this contract super generic. That way this can be migrated to simpler as well

contracts/croncat-factory/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "croncat-factory"
 version = "0.1.0"
-authors = []
+authors = ["buckram <buckram123@gmail.com>"]
```
TrevorJTClarke, line 4: k, another dumb idea = `{ workspace = true }` could work here? 
Im fine with individual here too

contracts/croncat-factory/src/contract.rs
```diff
@@ -13,19 +13,16 @@ use cw_utils::parse_reply_instantiate_data;
...
+    TEMP_REPLY,
 };
 
 // version info for migration info
 const CONTRACT_NAME: &str = "crates.io:croncat-factory";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-const CRONCAT_MANAGER_REPLY_ID: u64 = 0;
-const CRONCAT_TASKS_REPLY_ID: u64 = 1;
-const CRONCAT_AGENTS_REPLY_ID: u64 = 2;
```
TrevorJTClarke: TY!

contracts/croncat-factory/src/contract.rs
```diff
@@ -13,19 +13,16 @@ use cw_utils::parse_reply_instantiate_data;
...
-
 /// Save metadata and generate wasm msg
+/// Note: this will override contract metadata if same contract name and version was stored already
```
TrevorJTClarke, line 25: GTK, thanks!

contracts/croncat-factory/src/contract.rs
```diff
@@ -68,69 +69,9 @@ pub fn instantiate(
...
-        .collect::<StdResult<Vec<_>>>()?;
-
     Ok(Response::new()
         .add_attribute("action", "instantiate")
-        .add_attribute("sender", info.sender)
-        .add_submessage(croncat_manager_msg)
-        .add_submessage(croncat_tasks_msg)
-        .add_submessage(croncat_agents_msg)
-        .add_submessages(query_modules_msg))
+        .add_attribute("sender", info.sender))
```
TrevorJTClarke, line 74: Appreciate this removal!
For sender here, i believe its already part of the logs, so having that here is duplication

contracts/croncat-factory/src/contract.rs
```diff
@@ -155,45 +97,76 @@ pub fn execute(
...
-            new_changelog,
-        } => execute_update_metadata_changelog(deps, contract_name, version, new_changelog),
+            changelog_url,
+            schema,
+        } => execute_update_metadata(deps, contract_name, version, changelog_url, schema),
     }
 }
 
-fn execute_update_metadata_changelog(
+fn execute_update_config(deps: DepsMut, owner_addr: String) -> Result<Response, ContractError> {
```
TrevorJTClarke, line 109: if we're updating owner, we need to validate previous owner is the sender for safety

contracts/croncat-factory/src/contract.rs
```diff
@@ -155,45 +97,76 @@ pub fn execute(
...
     version: [u8; 2],
     new_changelog: Option<String>,
+    schema: Option<String>,
 ) -> Result<Response, ContractError> {
-    CONTRACT_METADATAS.update(deps.storage, (&contract_name, &version), |metadata_res| {
-        match metadata_res {
-            Some(mut metadata) => {
-                metadata.changelog_url = new_changelog;
-                Ok(metadata)
+    let metadata =
```
TrevorJTClarke, line 126: Need to check if owner is sender

contracts/croncat-factory/src/contract.rs
```diff
@@ -155,45 +97,76 @@ pub fn execute(
...
+        CONTRACT_METADATAS.update(deps.storage, (&contract_name, &version), |metadata_res| {
+            match metadata_res {
+                Some(mut metadata) => {
+                    // Update only if it contains values
+                    // No reason to set it from Some(x) to None
+                    if new_changelog.is_some() {
+                        metadata.changelog_url = new_changelog;
+                    }
+                    if schema.is_some() {
+                        metadata.schema = schema;
```
TrevorJTClarke, line 136: noice!

TrevorJTClarke (changes requested):
This is great @Buckram123!! Couple minor owner checks needed

contracts/croncat-factory/src/contract.rs
Buckram123: @TrevorJTClarke 
https://github.com/CronCats/cw-croncat/blob/b81b432ab363ef6752534d78a74c9ff482434723/contracts/croncat-factory/src/contract.rs#L84
It's checked for every execute action inside `factory`!

packages/croncat-sdk-factory/src/msg.rs
```diff
@@ -1,53 +1,85 @@
...
+    /// Update fields of the contract metadata
+    UpdateMetadata {
         contract_name: String,
         version: [u8; 2],
-        new_changelog: Option<String>,
+        changelog_url: Option<String>,
+        schema: Option<String>,
     },
 }
 // TODO: migrate
```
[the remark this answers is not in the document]
Buckram123: @TrevorJTClarke What do you mean by that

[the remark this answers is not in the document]
TrevorJTClarke: The migration logic :)

contracts/croncat-factory/src/contract.rs
TrevorJTClarke: oh yeah - thats right - sorry about that :)

TrevorJTClarke (approved):

packages/croncat-sdk-factory/src/msg.rs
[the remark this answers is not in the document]
Buckram123: I was looking at fields instead of `TODO` :duck:

PR #271, opened 2023-01-23
[FEATURE] Tasks split

Buckram123:
TBD:
- [ ] Tests
- [x] Comments


packages/croncat-sdk-core/src/lib.rs
```diff
@@ -1 +1,3 @@
+pub mod internal_messages;
```
Buckram123, line 1: Let's try to use messages wrappers to avoid importing our contracts as much as possible

PR #273, opened 2023-01-25
Schemas for query contracts

NinoLipartiia:


scripts/schema.sh
```diff
@@ -0,0 +1,22 @@
...
+cargo run --example schema > /dev/null
+
+# When all schemas are ready, can create schemas like this:
+# for f in ./contracts/*
+# do
+#   cd "$f"
+#   CMD="cargo run --example schema"
+#   cargo run --example schema
+#   cd "$START_DIR"
+# done
```
Buckram123, line 22: I'm ok with errors during schema generation for now, what do you think? @deveusss @mikedotexe @TrevorJTClarke

Buckram123 (approved):
LGTM

scripts/schema.sh
mikedotexe: yeah, for now it's fine, we mainly need the contracts to be ready for the audit. I also trust the schema issues will be much easier once we merge in the separate contracts that are being split out

mikedotexe (approved):
Thanks!

PR #274, opened 2023-01-26
Due to return macros, use allow unused_imports in a couple places

mikedotexe:
Cleaning up so we don't have this when running `just build`. Also making sure that `just test` works

Screen Shot 2023-01-26 at 10 44 36 AM

### DIFF
--- a/contracts/mod-balances/src/msg.rs
+++ b/contracts/mod-balances/src/msg.rs
@@ -1,5 +1,6 @@
 use crate::types::HasBalanceComparator;
 use cosmwasm_schema::{cw_serde, QueryResponses};
+#[allow(unused_imports)]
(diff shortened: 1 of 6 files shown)


contracts/mod-balances/src/msg.rs
```diff
@@ -1,5 +1,6 @@
 use crate::types::HasBalanceComparator;
 use cosmwasm_schema::{cw_serde, QueryResponses};
+#[allow(unused_imports)]
 use mod_sdk::types::QueryResponse;
```
Buckram123, line 4:
```suggestion
```
Hey, I prefer it to be full path, like this
```
    #[returns(mod_sdk::types::QueryResponse)]
```

PR #283, opened 2023-01-27
[FIX] Fix deployment js scripts

deveusss:


TrevorJTClarke (approved):

scripts/deployment/src/agents.ts
```diff
@@ -32,8 +32,9 @@ export class AgentClient {
                     "changelog_url": "https://github.com/croncats",
                     "schema": "",
                     "msg": Buffer.from(JSON.stringify({
-                        manager_addr: managerAddress,
-                        tasks_addr: tasksAddress
+                        "croncat_factory_addr": factoryAddress,
```
Buckram123:
```suggestion
```
Let's have this field be `sender` by default, we aren't planning to instantiate it in other ways, right? But if needed - can be changed inside `update_config`, thoughts? @TrevorJTClarke @mikedotexe

Buckram123 (changes requested):

Buckram123 (approved):

PR #285, opened 2023-01-27
Add just script/files to remove workspaces for faster builds and deploys

mikedotexe:
This branch tackles something unfortunate, which is that when we use Cargo workspaces, we gain the ability to not have a zillion `Cargo.toml` manifests with versions. But sadly it seems that right we lose our ability to "quickly build" all the contracts with cargo. We have to use `workspace-optimizer` but at this point, that takes a longer time than is reasonable.

If you do `cargo build` (with all the proper flags and `wasm32` stuff) it will create binaries that will not work when you try to deploy them.

To get around this, the manifests are symlinks to a new `manifests` directory that contain two versions of the manifest. One of them is for "regular" builds (aka `cargo build …`) and the other is for workspaces.

If you want to build, we can swap symlinks and build it. This is built into a new command in the `justfile`

    just cbuild

And then when it comes to deploying these guys, you can use:

    just cdeploy

What that does is set a simple environment variable in the deployment CosmJS code, setting `REGULAR="yes"` (silly, it's okay for now) And when the end-to-end testing code sees this value, it'll look for the contract binaries in the proper directory. And similarly, that same logic exists when running `go` (which is basically building TypeScript and running the output)

You can see details in the `package.json` file under the scripts:

- e2e-regular
- go-regular

---

**Note**: there may be confusion about the `Cargo.lock` files that might be tempting to revision. Let's just know that if your git is telling you about changes to the lockfiles, it might be due to workspaces/regular builds.


mikedotexe:
it'll be less distracting if i close this

scripts/deployment/package.json
```diff
@@ -7,9 +7,11 @@
     "postinstall": "cp .env.template .env",
     "build": "tsc",
     "start": "node build/index.js",
-    "go": "tsc && node build/index.js",
+    "go": "tsc && cross-env NODE_OPTIONS=--experimental-vm-modules NODE_NO_WARNINGS=1 node build/index.js",
     "go-regular": "tsc && REGULAR=\"yes\" node build/index.js",
-    "e2e": "tsc && node build/e2e.js",
+    "e2e": "tsc && cross-env NODE_OPTIONS=--experimental-vm-modules NODE_NO_WARNINGS=1 node build/e2e.js",
+    "e2e-regular": "tsc && cross-env NODE_OPTIONS=--experimental-vm-modules NODE_NO_WARNINGS=1 REGULAR=\"yes\" node build/e2e.js",
+    "mike": "tsc && node build/mike.js",
```
TrevorJTClarke, line 14: i dont see this script?

deveusss: I suppose its scripts/cargo-build.sh

TrevorJTClarke:
going to close this, as we will be moving to release CI/CD

PR #286, opened 2023-01-27
[FIX] Get_agent_tasks query method, remove additional args

deveusss:


contracts/croncat-agents/src/contract.rs
```diff
@@ -165,8 +158,7 @@ fn query_get_agent_ids(
 fn query_get_agent_tasks(
     deps: Deps,
     env: Env,
-    account_id: String,
```
mikedotexe, line 168: Don't we need this?

contracts/croncat-agents/src/contract.rs
```diff
@@ -165,8 +158,7 @@ fn query_get_agent_ids(
...
-    slots: (Option<u64>, Option<u64>), //block_slots,cron_slots
+    account_id: String, //block_slots,cron_slots
```
mikedotexe, line 161:
```suggestion
    account_id: String,
```

PR #289, opened 2023-01-30
[FEATURE] Proxy call implementation

Buckram123:
~~WIP:~~

- [x] - tests

Changes, from original implementation:

1. `reply_queue` and `reply_index` was replaced with a single 
```
REPLY_QUEUE: Item<QueueItem>
 ```
2. `QueueItem` got simplyfied from https://github.com/CronCats/cw-croncat/blob/b920d1a5d5b1730397b04acdbba205eacfd487bd/contracts/cw-croncat/src/state.rs#L62-L76
to:
```rust
pub struct QueueItem {
    pub task: croncat_sdk_tasks::types::TaskResponse,
    pub agent_addr: Addr,
    /// Storing any errors that happened to return
    pub failures: Vec<(u8, String)>,
}
```
3. Submessages of the task's actions only `reply_on_error`, except for the last submessage, where we finalize task

Reasoning for 1 and 2:
- We don't need to keep such complex state if cosmwasm protects us from reentrancy attacks by design
  - source: https://docs.cosmwasm.com/docs/architecture/smart-contracts/#avoiding-reentrancy-attacks

Reasoning for 3:
- Cosmwasm keeps ordering of the submessages and we don't need to do anything with successful events, unless it's a last action
  - sources: https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.Response.html#structfield.messages and https://github.com/CosmWasm/cosmwasm/blob/main/SEMANTICS.md#order-and-rollback


contracts/croncat-manager/src/contract.rs
```diff
@@ -164,25 +157,129 @@ fn execute_proxy_call(
...
+    }) {
+        return Err(ContractError::NoTaskForAgent {});
+    }
+
+    // Get a task
+    let tasks_addr = get_tasks_addr(&deps.querier, &config)?;
+    let current_task: Option<croncat_sdk_tasks::types::TaskResponse> =
+        deps.querier.query_wasm_smart(
+            tasks_addr,
+            &croncat_sdk_tasks::msg::TasksQueryMsg::TaskWithTransforms { task_hash },
```
TrevorJTClarke, line 237: @Buckram123 why `TaskWithTransforms` here?

contracts/croncat-manager/src/contract.rs
```diff
@@ -164,25 +157,129 @@ fn execute_proxy_call(
...
+    {
+        // Task is no longer valid
+        let coins_transfer = remove_task_balance(
+            deps.storage,
+            task_balance,
+            &task.owner_addr,
+            &config.native_denom,
+            task.task_hash.as_bytes(),
+        )?;
+        // TODO: help to find a better attributes for this
```
TrevorJTClarke, line 257: seems enough for me. less is more :)

contracts/croncat-manager/src/contract.rs
```diff
@@ -347,7 +443,50 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+                Ok(Response::new().add_message(bank_send))
+            } else {
+                Ok(Response::new())
+            }
+        }
+        _ => {
+            let mut queue_item = REPLY_QUEUE.load(deps.storage)?;
+            let last = parse_reply_msg(deps.storage, &mut queue_item, msg);
+            if last {
+                let failures: Vec<Attribute> = queue_item
```
TrevorJTClarke, line 480: Does this mean we'll be able to debug submsgs!?

contracts/croncat-manager/src/contract.rs
Buckram123: It queries task with queries, and if it has transforms it applies them.
What do you think, should transforms happen on the tasks or manager contract?

contracts/croncat-manager/src/contract.rs
Buckram123: Yeah, so anyone can check his executed task and see what actions failed!)

contracts/croncat-manager/src/contract.rs
TrevorJTClarke: are you saying that it does queries & transforms within the tasks contract?
I figured all proxy_call logic including queries / transforms would live within manager. This way its consistent behaviour across tasks (in the event we support multiple task contracts 😯)

contracts/croncat-manager/src/contract.rs
TrevorJTClarke: This is life saving then!!!!!!!!

contracts/croncat-manager/src/contract.rs
Buckram123: Okay then, gonna move it back to the manager then 🧨

contracts/croncat-manager/src/contract.rs
```diff
@@ -246,17 +246,43 @@ fn execute_proxy_call_with_queries(
...
+            .into(),
+        )?;
+        if !query_res.result {
+            return Err(ContractError::TaskNotReady {});
+        }
+        query_responses.push(query_res.data);
+    }
+    replace_values(&mut task, query_responses)?;
+
+    let invalidated_after_transform = if let Ok(amounts) =
```
TrevorJTClarke, line 273: thats pretty neat!

contracts/croncat-manager/src/contract.rs
```diff
@@ -246,17 +246,43 @@ fn execute_proxy_call_with_queries(
...
+    let mut query_responses = Vec::with_capacity(task.queries.as_ref().unwrap().len());
+    for query in task.queries.iter().flatten() {
+        let query_res: mod_sdk::types::QueryResponse = deps.querier.query(
+            &WasmQuery::Smart {
+                contract_addr: query.query_mod_addr.clone(),
+                msg: query.msg.clone(),
+            }
+            .into(),
+        )?;
+        if !query_res.result {
```
TrevorJTClarke, line 266: Still thinking this piece through. I was hoping we support a `match` so we can do just regular query or CronCat flavored `QueryResponse` which does the check for result here.
This allows any task to have a query + transform for both schedule or event driven. Where only event driven ones have `QueryResponse` result check.

HMU in discord if you wanna clarify!

TrevorJTClarke (changes requested):
Looking like solid progress!!
Left a couple notes in there @Buckram123 :)

contracts/croncat-manager/src/contract.rs
Buckram123: @TrevorJTClarke so adding `check_result` toggle to the CronCatQuery is enough right?

TrevorJTClarke: @Buckram123 yeah - thats a good plan

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -59,6 +59,12 @@ pub struct TaskRequest {
     pub transforms: Option<Vec<Transform>>,
 
     /// How much of cw20 coin is attached to this task
+    /// This will be taken from the manager's contract temporary "Users balance"
+    /// and attached directly to the task's balance.
+    /// 
+    /// Note: Unlike other coins ( which get refunded to the task creator in the same transaction as task removal)
+    /// cw20's will get moved back to the temporary "Users balance".
+    /// This is done primarily to save up gas from executing another contract during `proxy_call`
```
Buckram123, line 67: > Cw20 during create_task - how does this work? When i pass in task.cw20, is it a reference to a pre-deposited balance or is it part of a receiver chain?

Hey @TrevorJTClarke Is it good enough description about lifetime of the cw20's inside a task?

TrevorJTClarke: thats perfect, thanks!

contracts/croncat-manager/src/contract.rs
```diff
@@ -270,13 +271,14 @@ fn execute_proxy_call_with_queries(
             }
             .into(),
         )?;
-        if !query_res.result {
+        if query.check_result && !query_res.result {
```
TrevorJTClarke, line 274: This feels much better. I can't think of how to do better :P

contracts/croncat-manager/src/contract.rs
```diff
@@ -285,7 +287,8 @@ fn execute_proxy_call_with_queries(
     } else {
         true
     };
-    // Need to re-check if task has enough cw20's, cause it could have been changed through transform
+    // Need to re-check if task has enough cw20's
+    // because it could have been changed through transform
```
TrevorJTClarke, line 291: This is great - actually had a note from Pupmos months ago on this exact problem happening from not checking

TrevorJTClarke (changes requested):
@Buckram123 tiny tiny thangs

contracts/croncat-manager/src/tests/contracts.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused)]
```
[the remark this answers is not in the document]
Buckram123: Left it on purpose, have few not used (anymore) contracts for testing. If we find some strange behavior or think about some task-flow that needs to be covered would be nice to get it ready for fast-checking it

contracts/croncat-manager/src/tests/helpers.rs
```diff
@@ -218,6 +224,45 @@ pub(crate) fn init_mod_balances(app: &mut App, factory_addr: &Addr) -> Addr {
     metadata.unwrap().contract_addr
 }
 
+#[allow(unused)]
```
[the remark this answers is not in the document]
Buckram123: Maybe used later. It is in the tests, should be fine to keep it unused)

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -285,6 +291,7 @@ impl Task {
 pub struct CroncatQuery {
     pub query_mod_addr: String,
     pub msg: Binary,
+    pub check_result: bool,
```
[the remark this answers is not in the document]
Buckram123: So change to the `contract_addr` and add comment to what type of addr it is, right?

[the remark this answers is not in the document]
TrevorJTClarke: yes please!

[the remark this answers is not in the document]
TrevorJTClarke: `QueryResponse` is still the response tho, not the request. so it shouldnt change

contracts/croncat-manager/src/contract.rs
Buckram123: Lol :D

TrevorJTClarke (approved):

PR #294, opened 2023-01-31
Beta/split 0.0.0 fixes

deveusss:
- [x] In query_users_balances change "wallet" to "address" to remove confusion
- [x] In instantiate TREASURY_BALANCE starts as zero, but info could have funds - so maybe add funds into TREASURY_BALANCE or just error if info funds attached
- [x] Can not pass funds in factory deploy

### DIFF
--- a/contracts/croncat-agents/src/contract.rs
+++ b/contracts/croncat-agents/src/contract.rs
@@ -3,6 +3,7 @@ use crate::error::ContractError;
 use crate::external::*;
 use crate::msg::*;
 use crate::state::*;
+use cosmwasm_std::Uint64;
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::{
     entry_point, has_coins, to_binary, Addr, Binary, Coin, Deps, DepsMut, Env, MessageInfo,
@@ -160,11 +161,7 @@ fn query_get_agent_ids(
     Ok(GetAgentIdsResponse { active, pending })
 }
 
-fn query_get_agent_tasks(
-    deps: Deps,
-    env: Env,
-    account_id: String,
-) -> StdResult<Option<AgentTaskResponse>> {
+fn query_get_agent_tasks(deps: Deps, env: Env, account_id: String) -> StdResult<AgentTaskResponse> {
     let account_id = deps.api.addr_validate(&account_id)?;
     let active = AGENTS_ACTIVE.load(deps.storage)?;
     if !active.contains(&account_id) {
@@ -176,7 +173,10 @@ fn query_get_agent_tasks(
 
     let (block_slots, cron_slots) = croncat_tasks_contract::query_tasks_slots(deps, &config)?;
     if block_slots == 0 && cron_slots == 0 {
-        return Ok(None);
+        return Ok(AgentTaskResponse {
+            num_cron_tasks: Uint64::zero(),
+            num_block_tasks: Uint64::zero(),
+        });
     }
     AGENT_TASK_DISTRIBUTOR
         .get_agent_tasks(
--- a/contracts/croncat-agents/src/contract.rs
+++ b/contracts/croncat-agents/src/contract.rs
@@ -410,7 +410,7 @@ fn unregister_agent(
     let mut active_agents: Vec<Addr> = AGENTS_ACTIVE.load(deps.storage)?;
     if let Some(index) = active_agents.iter().position(|addr| addr == agent_id) {
         //Notify the balancer agent has been removed, to rebalance itself
-        AGENT_TASK_DISTRIBUTOR.on_agent_unregistered(deps.storage, agent_id)?;
+        AGENT_STATS.remove(deps.storage, agent_id);
         active_agents.remove(index);
         AGENTS_ACTIVE.save(deps.storage, &active_agents)?;
     } else {
--- a/contracts/croncat-agents/src/distributor.rs
+++ b/contracts/croncat-agents/src/distributor.rs
@@ -18,12 +18,6 @@ pub trait RoundRobinAgentTaskDistributor<'a> {
         agent_id: Addr,
         slot_items: (Option<u64>, Option<u64>),
     ) -> Result<AgentTaskResponse, ContractError>;
-    #[doc = r"Removes balancer stats and rebalances"]
-    fn on_agent_unregistered(
-        &self,
-        storage: &'a mut dyn Storage,
-        agent_id: &Addr,
-    ) -> Result<(), ContractError>;
 
     #[doc = r"Updates agent stats when agent completed task on specified slot"]
     fn on_task_completed(
@@ -121,19 +115,6 @@ impl<'a> RoundRobinAgentTaskDistributor<'a> for AgentTaskDistributor {
             num_block_tasks,
             num_cron_tasks,
         })
-    }
-
-    fn on_agent_unregistered(
-        &self,
-        storage: &'a mut dyn Storage,
-        agent_id: &Addr,
-    ) -> Result<(), ContractError> {
-        let active = AGENTS_ACTIVE.load(storage)?;
-        if !active.contains(agent_id) {
-            return Err(ContractError::AgentNotRegistered);
-        }
-        AGENT_STATS.remove(storage, agent_id);
-        Ok(())
     }
 
     fn on_task_completed(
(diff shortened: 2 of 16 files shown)


contracts/croncat-manager/src/contract.rs
```diff
@@ -63,6 +63,14 @@ pub fn instantiate(
         .transpose()?
         .unwrap_or_else(|| info.sender.clone());
 
+    let treasury_funds = info
+        .funds
+        .iter()
+        .find(|coin| coin.denom == denom)
+        .ok_or(ContractError::NoTreasuryFundsAttached {})?;
```
Buckram123, line 70: Let's say we have this situation: 
- [20coinA, 30coinB] attached and `config.denom` is `coinA`

Should we fail tx if other coins were attached or what to do with coinB? @TrevorJTClarke

Buckram123, line 70:
```suggestion
```
It should allow empty deposit!

contracts/croncat-manager/src/tests/contract_tests.rs
```diff
@@ -361,7 +369,9 @@ fn cw20_bad_messages() {
     let mut app = default_app();
 
     let instantiate_msg: InstantiateMsg = default_instantiate_message();
-    let manager_addr = init_manager(&mut app, instantiate_msg, &[]).unwrap();
+    let send_funds: &[Coin] = &[coin(600, instantiate_msg.denom.clone())];
+
+    let manager_addr = init_manager(&mut app, instantiate_msg, send_funds).unwrap();
```
Buckram123, line 374: Let's not have it standard way to initialize contract

Buckram123 (changes requested):
I personally against attached deposit being the standard way to initialize treasury balance, it doesn't make sense if croncat-factory owner is a croncat-dao itself, why would they ping-pong balance in and out?

Can't think it can be useful for anything, so I'm more towards blocking attached funds on initialize, at least for now! What's your thoughts on that? @TrevorJTClarke  @mikedotexe @NinoLipartiia @deveusss

contracts/croncat-manager/src/contract.rs
deveusss: Trevors comment "TREASURY_BALANCE or just error if no funds attached"

contracts/croncat-manager/src/tests/contract_tests.rs
deveusss: Copy it from agent tests

contracts/croncat-manager/src/contract.rs
deveusss: For now its not failing if others are attached by if no funds with default denom - it will fail

Buckram123: "or just error if info funds attached"
pretty sure it means
```rust
if info.funds.is_empty{
    return Err();
}
```

contracts/croncat-manager/src/tests/contract_tests.rs
Buckram123: I mean about positive balance at treasury from the start)

contracts/croncat-manager/src/contract.rs
deveusss: ahh, true

deveusss: updated it, so now funds are added into treasure balance

contracts/croncat-manager/src/balances.rs
```diff
@@ -230,7 +230,7 @@ pub fn execute_refill_native_balance(
         .may_load(deps.storage, task_hash.as_bytes())?
         .ok_or(ContractError::NoTaskHash {})?;
 
-    if info.funds.len() > 2 {
+    if info.funds.len() > 3 {
```
TrevorJTClarke, line 233: @deveusss @Buckram123 whats the hardcoded value here for? Is there a reason its not higher like 10 or just simply 1?

justfile
```diff
@@ -14,8 +14,8 @@ build:
 	cargo build --release --lib --target wasm32-unknown-unknown
 deploy:
 	#!/bin/bash
-	chmod +x ./scripts/testnet/deploy.sh
-	./scripts/testnet/deploy.sh -w # only wasm update
+	cd ./scripts/deployment
+	yarn go
```
TrevorJTClarke, line 18: @deveusss whats the core use-case here? Will CI/CD be running this? If so, can we guarantee other things like build & optimize run first?

scripts/deployment/src/manager.ts
```diff
@@ -20,6 +20,7 @@ export class ManagerClient {
 		const codeId = uploadRes.codeId
 		const checksums = await getChecksums()
 		const githash = await getGitHash()
+		const funds:Coin[]=[coin(50,denom)];
```
TrevorJTClarke, line 23: @deveusss likely need some `cargo fmt --` love here

TrevorJTClarke (approved):
@deveusss lookin good! couple minor things.

RE: Things @Buckram123 brought up - theres some back in forth in my head about funds for future needs, i would rather support this way to start and remove if need be.

contracts/croncat-manager/src/balances.rs
Buckram123: Didn't we decide to keep up to 1 `config.denom` and 1 non-`config.denom`(ibc) same with up to 1 cw20?

Buckram123: It should be here then, or limitless if you think it's a bad idea

deveusss:
> @deveusss lookin good! couple minor things.
> 
> RE: Things @Buckram123 brought up - theres some back in forth in my head about funds for future needs, i would rather support this way to start and remove if need be.

So we are supporting passing funds in instantiate?

deveusss:
> I personally against attached deposit being the standard way to initialize treasury balance, it doesn't make sense if croncat-factory owner is a croncat-dao itself, why would they ping-pong balance in and out?
> 
> Can't think it can be useful for anything, so I'm more towards blocking attached funds on initialize, at least for now! What's your thoughts on that? @TrevorJTClarke @mikedotexe @NinoLipartiia @deveusss

Whats wrong with passing funds into your contract at init? why not if we have this possibility?

justfile
deveusss: this is not for CI/CD but for deploy on testnet and check things in testnet

scripts/deployment/src/manager.ts
deveusss: √

Buckram123 (approved):

contracts/croncat-agents/src/contract.rs
```diff
@@ -176,7 +173,10 @@ fn query_get_agent_tasks(
 
     let (block_slots, cron_slots) = croncat_tasks_contract::query_tasks_slots(deps, &config)?;
     if block_slots == 0 && cron_slots == 0 {
-        return Ok(None);
+        return Ok(AgentTaskResponse {
+            num_cron_tasks: Uint64::zero(),
+            num_block_tasks: Uint64::zero(),
+        });
```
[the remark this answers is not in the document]
deveusss: this is old code migrated form old contract)))

contracts/croncat-agents/src/contract.rs
```diff
@@ -410,7 +410,7 @@ fn unregister_agent(
     let mut active_agents: Vec<Addr> = AGENTS_ACTIVE.load(deps.storage)?;
     if let Some(index) = active_agents.iter().position(|addr| addr == agent_id) {
         //Notify the balancer agent has been removed, to rebalance itself
-        AGENT_TASK_DISTRIBUTOR.on_agent_unregistered(deps.storage, agent_id)?;
+        AGENT_STATS.remove(deps.storage, agent_id);
```
TrevorJTClarke, line 413: Thank you!

contracts/croncat-agents/src/distributor.rs
```diff
@@ -104,33 +101,20 @@ impl<'a> RoundRobinAgentTaskDistributor<'a> for AgentTaskDistributor {
                 }
                 let agent_tasks_total = total_tasks.saturating_div(agent_count) + extra;
 
-                Ok((agent_tasks_total.into(), extra.into()))
+                Ok(agent_tasks_total.into())
```
TrevorJTClarke, line 104: So much better!!

TrevorJTClarke (approved):

PR #295, opened 2023-01-31
Ft/factory relayer

TrevorJTClarke:


contracts/croncat-factory/src/contract.rs
```diff
@@ -44,6 +44,8 @@ fn init_save_metadata_generate_wasm_msg(
         &metadata,
     )?;
     LATEST_VERSIONS.save(storage, &init_info.contract_name, &init_info.version)?;
+
+    // TODO: Once cosmwasm 1.2 is available, change to Instantiate2 https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/results/cosmos_msg.rs#L169
```
Buckram123, line 48: I think it's available, but juno have not updated to it yet. Not sure about other chains

contracts/croncat-factory/src/contract.rs
```diff
@@ -118,6 +121,43 @@ fn execute_update_config(deps: DepsMut, owner_addr: String) -> Result<Response,
...
 }
 
+fn execute_proxy(
+    deps: DepsMut,
+    info: MessageInfo,
+    msg: WasmMsg,
+) -> Result<Response, ContractError> {
+    let config = &CONFIG
+        .may_load(deps.storage)?
+        .ok_or(ContractError::UnknownContract {})?;
```
Buckram123, line 131:
```suggestion
    let config = CONFIG.load(deps.storage)?;
```
It should load and no need to take a reference here)

contracts/croncat-factory/src/contract.rs
```diff
@@ -118,6 +121,43 @@ fn execute_update_config(deps: DepsMut, owner_addr: String) -> Result<Response,
...
+    // Only accept WasmMsg::Execute
+    let contract_addr = match &msg {
+        WasmMsg::Execute {
+            contract_addr,
+            funds: _,
+            msg: _,
+        } => contract_addr,
+        // Disallow unknown messages
+        _ => {
+            return Err(ContractError::Unauthorized {});
```
Buckram123, line 147: Let's have a clearer error

contracts/croncat-factory/src/contract.rs
```diff
@@ -118,6 +121,43 @@ fn execute_update_config(deps: DepsMut, owner_addr: String) -> Result<Response,
...
+        }
+    };
+
+    // Only allow msgs that have existing contract versions
+    CONTRACT_ADDRS_LOOKUP
+        .may_load(deps.storage, deps.api.addr_validate(&contract_addr)?)?
+        .ok_or(ContractError::UnknownContract {})?;
```
Buckram123, line 154: Consider using `has()` here instead, if not planning to do anything with a metadata
https://docs.rs/cw-storage-plus/latest/cw_storage_plus/struct.Map.html#method.has

contracts/croncat-factory/src/contract.rs
```diff
@@ -171,6 +212,7 @@ fn execute_remove(
 
     CONTRACT_METADATAS.remove(deps.storage, (&contract_name, &version));
     CONTRACT_ADDRS.remove(deps.storage, (&contract_name, &version));
+    CONTRACT_ADDRS_LOOKUP.remove(deps.storage, contract_addr);
```
Buckram123, line 215: Nice

contracts/croncat-factory/src/state.rs
```diff
@@ -13,6 +13,8 @@ pub const CONTRACT_METADATAS: Map<(&str, &[u8]), ContractMetadata> = Map::new("c
 
 /// Contract name with the version to the Addr
 pub const CONTRACT_ADDRS: Map<(&str, &[u8]), Addr> = Map::new("contract_addrs");
+/// Contract Addr linked to contract name
+pub const CONTRACT_ADDRS_LOOKUP: Map<Addr, String> = Map::new("contract_addrs_lookup");
```
Buckram123, line 17: Sounds very useful, let's add query method to get these!

contracts/croncat-factory/src/contract.rs
```diff
@@ -118,6 +121,43 @@ fn execute_update_config(deps: DepsMut, owner_addr: String) -> Result<Response,
...
+    }
+
+    // Only accept WasmMsg::Execute
+    let contract_addr = match &msg {
+        WasmMsg::Execute {
+            contract_addr,
+            funds: _,
+            msg: _,
+        } => contract_addr,
+        // Disallow unknown messages
```
Buckram123, line 145: Is there any scenario you can think of we can use that to `instantiate`/`migrate`?

Buckram123 (approved):
Looks pretty good! Thanks for adding `contract_addrs_lookup`!

contracts/croncat-factory/src/contract.rs
TrevorJTClarke: I want migrate to be explicitly different. :)

contracts/croncat-factory/src/state.rs
TrevorJTClarke: we return the contract_addr within the version query response already, this can remain simple lookup

contracts/croncat-factory/src/contract.rs
TrevorJTClarke: good call!

contracts/croncat-factory/src/contract.rs
TrevorJTClarke: sure

contracts/croncat-factory/src/contract.rs
TrevorJTClarke: nice!

contracts/croncat-factory/src/contract.rs
TrevorJTClarke: yeah - waiting for future :)

contracts/croncat-factory/src/contract.rs
Buckram123: > I want migrate to be explicitly different. :)

Right, and `deploy` is our `instantiate`

contracts/croncat-factory/src/state.rs
Buckram123: > we return the contract_addr within the version query response already, this can remain simple lookup

But we don't return `contract_name` from the `addr`, in any case, if someone wants to check if "this addr" is in our state) Maybe some form of trust-check, wdyt? Or it should be allowed only by the raw queries?

TrevorJTClarke: i guess thats a good consideration, lemme think about it

contracts/croncat-factory/src/contract.rs
```diff
@@ -378,5 +418,7 @@ pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractE
         &contract_address,
     )?;
 
+    CONTRACT_ADDRS_LOOKUP.save(deps.storage, contract_address, &contract_name)?;
```
mikedotexe, line 421: This is sweet.

mikedotexe (approved):

contracts/croncat-factory/src/tests/factory_tests.rs
```diff
@@ -268,8 +270,13 @@ fn deploy_check() {
         )
         .unwrap();
     assert_eq!(tasks_metadatas.len(), 1);
-    let _tasks_metadata = tasks_metadatas.remove(0);
-    // TODO check it's tasks
+    let tasks_metadata = tasks_metadatas.remove(0);
+    // check it's tasks
```
mikedotexe, line 274:
```suggestion
    // check its tasks
```
🤓

contracts/mod-dao/src/tests/passed_proposals.rs
```diff
@@ -345,6 +345,7 @@ fn test_multiple_check_passed_proposals() {
             &dao_proposal_multiple::msg::ExecuteMsg::Vote {
                 proposal_id: 2 * num as u64,
                 vote: MultipleChoiceVote { option_id: 0 },
+                rationale: Some("because_pulled_pork_mac_n_cheese".to_string()),
```
mikedotexe, line 348: 👏🏼

mikedotexe: oh wait, was this `rationale` the thing that was updated on the DAO DAO end?

contracts/croncat-factory/src/error.rs
```diff
@@ -16,6 +16,9 @@ pub enum ContractError {
     #[error("Unknown contract name")]
     UnknownContract {},
 
+    #[error("Unknown contract method")]
```
mikedotexe, line 19:
```suggestion
    #[error("Unknown or unsupported contract method")]
```

contracts/croncat-factory/src/contract.rs
```diff
@@ -121,6 +123,41 @@ fn execute_update_config(deps: DepsMut, owner_addr: String) -> Result<Response,
...
+    };
+
+    // Only allow msgs that have existing contract versions
+    if !CONTRACT_ADDRS_LOOKUP.has(deps.storage, deps.api.addr_validate(contract_addr)?) {
+        return Err(ContractError::UnknownContract {});
+    }
+
+    Ok(Response::new()
+        .add_attribute("action", "proxy")
+        .add_message(msg))
```
mikedotexe, line 158: The only bummer with using a message vs submessage is the person may have a rough time seeing what went wrong. I think it's possible for us to retrieve/return error messages that may have occurred from the proxy message, by using `reply`.

Looks like the factory already has a `reply` but it's only being used to see how contract instantiation went.

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -343,7 +343,7 @@ pub struct SlotIdsResponse {
     pub block_ids: Vec<u64>,
 }
 
-#[derive(Debug, PartialEq)]
+#[cw_serde]
```
Buckram123, line 346:
```suggestion
#[allow(clippy::derive_partial_eq_without_eq)]
#[derive(Debug, PartialEq)]
```
Will work too, if you want to min-max wasm size

TrevorJTClarke: I saw that - i think we can do optimization path separate across all contracts

PR #296, opened 2023-02-01
Beta/split 0.0.0 tick action

deveusss:
- [x] Tick task - how can we standardize (tick for all contracts, triggered at manager level?)
	- [x] Only impl on agent contract, for cleaning up bad agents & nominating new
	- [x] execute_tick


scripts/deployment/src/manager.ts
```diff
@@ -20,6 +20,7 @@ export class ManagerClient {
 		const codeId = uploadRes.codeId
 		const checksums = await getChecksums()
 		const githash = await getGitHash()
+		const funds:Coin[]=[coin(50,denom)];
```
mikedotexe, line 23: What is this for?

mikedotexe (changes requested):

scripts/deployment/src/manager.ts
```diff
@@ -49,7 +50,7 @@ export class ManagerClient {
 			}
 		}
 
-		const instRes = await this.client.execute(sender, factoryAddress, deployMsg, executeGas);
+		const instRes = await this.client.execute(sender, factoryAddress, deployMsg, executeGas,null,funds);
```
mikedotexe, line 53: This line is breaking things, actually, it no longer outputs a correct contract address for the manager contract

Screen Shot 2023-01-31 at 7 00 42 PM

contracts/croncat-agents/src/contract.rs
```diff
@@ -602,6 +623,37 @@ fn agents_to_let_in(max_tasks: &u64, num_active_agents: &u64, total_tasks: &u64)
         0
     }
 }
+pub fn execute_tick(deps: DepsMut, env: Env) -> Result<Response, ContractError> {
+    let current_slot = env.block.height;
```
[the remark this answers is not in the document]
TrevorJTClarke: @mikedotexe dont change to height - the main reason is because we have 2 contexts, block based & time based -- height is restricted to block context and confuses all the things

scripts/deployment/src/manager.ts
TrevorJTClarke: woah - thats very weird & unexpected. Def needs to get removed

contracts/croncat-agents/src/contract.rs
```diff
@@ -394,61 +411,62 @@ fn accept_nomination_agent(
 /// Withdraws all reward balances to the agent payable account id.
 /// In case it fails to unregister pending agent try to set `from_behind` to true
 fn unregister_agent(
-    deps: DepsMut,
+    storage: &mut dyn Storage,
```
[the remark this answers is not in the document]
TrevorJTClarke: agree

contracts/croncat-agents/src/state.rs
```diff
@@ -18,3 +18,4 @@ pub const AGENTS_ACTIVE: Item<Vec<Addr>> = Item::new("agents_active");
 pub const AGENTS_PENDING: Deque<Addr> = Deque::new("agents_pending");
 pub const AGENT_STATS: Map<&Addr, AgentStats> = Map::new("agent_stats");
 pub const AGENT_TASK_DISTRIBUTOR: AgentTaskDistributor = AgentTaskDistributor::new();
+pub const DEFAULT_AGENTS_EVECT_THRESHOLD: u64 = 600;
```
TrevorJTClarke, line 21: `EVECT` - did you mean `EVICT`?

contracts/croncat-agents/src/contract.rs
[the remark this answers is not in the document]
TrevorJTClarke: Ah i see what you're saying @mikedotexe  - yeah, lets clear that up

contracts/croncat-agents/src/contract.rs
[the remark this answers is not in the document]
deveusss: there is a reason, unregister_agent is call in tick, so move occurs, but passing storage resolves that issue)

contracts/croncat-agents/src/contract.rs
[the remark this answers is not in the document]
deveusss: fixed

scripts/deployment/src/manager.ts
deveusss: coins were mandatory, so we updated manager, and coins are not mandatory

contracts/croncat-agents/src/state.rs
deveusss: oops) EJECT

scripts/deployment/src/manager.ts
deveusss: it was denom error, validator does not accept ustart in info.funds

deveusss: fixed

scripts/deployment/src/manager.ts
deveusss: fixed

mikedotexe (changes requested):
Hey @deveusss I see this is in the `UpdateConfig` but not the `InstantiateMsg`. Can we add the `agent_nomination_duration` there, too? And have it save during instantiate. Will be good to keep the initial instantiate message and the updating to be consistent)

mikedotexe:
I noticed that I am unable to `just build`, `just optimize`, `just deploy` without getting this error:

```
Error: Error when broadcasting tx 0AB8BCC6F9B5A9FA87B1BEA7B020440904F62F66E364DAE96E03BBBFA3E6CF90 at height 3232704. Code: 2; Raw log: failed to execute message; message index: 0: Error calling the VM: Error during static Wasm validation: Wasm contract doesn't have required export: "instantiate". Exports required by VM: ["allocate", "deallocate", "instantiate"].: create wasm contract failed
```

I'm not positive it's this branch, and am checking the BS branch)), but it's taking forever, but thought I'd share here before I forget.

Screen Shot 2023-02-01 at 9 16 51 PM

deveusss:
> agent_nomination_duration



> Hey @deveusss I see this is in the `UpdateConfig` but not the `InstantiateMsg`. Can we add the `agent_nomination_duration` there, too? And have it save during instantiate. Will be good to keep the initial instantiate message and the updating to be consistent)

hey @mikedotexe its in init too)

https://github.com/CronCats/cw-croncat/blob/beta/split-0.0.0-tick-action/contracts/croncat-agents/src/contract.rs#L52

PR #298, opened 2023-02-01
[FEATURE] paused check for non-library contract before removal inside factory

Buckram123:


contracts/croncat-factory/src/contract.rs
```diff
@@ -167,8 +168,21 @@ fn execute_remove(
 
     // Can't remove unpaused contract if not a library
     if metadata.kind != VersionKind::Library {
-        // Check if paused
-        todo!();
+        // universal paused checker
+        #[derive(Deserialize)]
+        #[serde(crate = "cosmwasm_schema::serde")]
+        struct CheckIfConfigIsPaused {
```
TrevorJTClarke, line 174: @Buckram123 why inline? Wouldn't this be better in packages?

TrevorJTClarke (approved):
awesome @Buckram123!!

gunna have merge conflicts with my branch - but ill deal with it

contracts/croncat-factory/src/contract.rs
Buckram123: > @Buckram123 why inline? Wouldn't this be better in packages?

It feels like it is the only place it can be used, so didn't feel to move it anywhere else. This way this type is only available in this scope and, for example, it won't be suggested by the IDE. 

Not sure what's the best practice here :(
Want to make it `pub`?

TrevorJTClarke: Aren't we doing this same thing on multiple contracts tho?

Buckram123: > Aren't we doing this same thing on multiple contracts tho?

Pretty sure it's the only place we have `deseralize-only-one-field` struct

mikedotexe: I feel like I'd rather be consistent and discoverable. I can already see myself in 6 months being like, "where is this thing coming from?" and then realizing we discussed this as an exception

PR #299, opened 2023-02-01
Fix test dealing with doc-tests

mikedotexe:
Teeny tiny thing that'll make `just test` pass. It was a comment with code that didn't follow:
https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments

### DIFF
--- a/packages/croncat-sdk-tasks/src/types.rs
+++ b/packages/croncat-sdk-tasks/src/types.rs
@@ -294,7 +294,8 @@ pub struct CroncatQuery {
     ///
     /// One requirement for custom contracts: query return value should be formatted as:
     /// ```
-    /// {
+    /// use cosmwasm_std::Binary;
+    /// struct ReturnValue {
     ///     result: bool,
     ///     data: Binary,
     /// }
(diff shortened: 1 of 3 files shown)


Buckram123 (approved):
This will also work, I'm ok with any
```
```no_run
```

Buckram123:
> ```
> ```no_run
> ```

oops, i meant ignore `instead` of `no_run` haha

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -292,13 +292,8 @@ pub struct CroncatQuery {
...
     ///
-    /// One requirement for custom contracts: query return value should be formatted as:
-    /// ```
-    /// {
-    ///     result: bool,
-    ///     data: Binary,
-    /// }
-    /// ```
+    /// One requirement for custom contracts: query return value should be formatted as a:
+    /// [`QueryResponse`]: mod_sdk::types::QueryResponse
```
Buckram123, line 296: Nice, I like it way more, but should be 
```suggestion
    /// [`mod_sdk::types::QueryResponse`]
```

Buckram123 (approved):

PR #301, opened 2023-02-01
for ts, add Nullable_* where needed (tasks and agents)

mikedotexe:
This one fixes stuff that breaks. Let's get ONE green light from CI, lul


mikedotexe:
@deveusss do you know why these TypeScript files are having problems where this `Nullable_` is needed with the underscore. Is this something you have an instinct for, after being in the TypeScript stuff a bit lately? I think you understand this better than me)

deveusss:
this one fixed, I think can be closed

mikedotexe:
Fixed by #304

PR #302, opened 2023-02-01
fix QueryResponse doc comment to actually link there

mikedotexe:
Per @Buckram123 noticing that it 

# DIDN'T ACTUALLY WORK, Mike!

I thought I might fix my work.


Buckram123 (approved):
Clean AF🤑

PR #304, opened 2023-02-01
fix: typegen and reponse types to fix issue with typescript

deveusss:
This PR is fixing typescript codegen


mikedotexe (approved):
Thank you, nice little trick of how DAO DAO also does it.

PR #306, opened 2023-02-02
Disabling feature "library" on the factory

Buckram123:
What happened: we imported `croncat-factory` with a "library" feature from `manager`, `agents` and `tasks` contract. And it made `croncat-factory` build it with a "library" feature by default, which is obviously wrong


deveusss:
good fix)

mikedotexe:
Yep, tested it and it fixed it, amazing thank you

PR #307, opened 2023-02-02
Tests for WithdrawAgentRewards

NinoLipartiia:


contracts/croncat-manager/src/balances.rs
```diff
@@ -228,7 +228,7 @@ pub fn execute_user_withdraw(
         .add_messages(msgs))
 }
 
-/// Execute: MoveBalances
+/// Execute: OwnerWithdraw
 /// Used by owner of the contract to move balances from the manager to treasury or owner address
```
Buckram123, line 232: Thanks!

contracts/croncat-manager/src/tests/manager_tests.rs
```diff
@@ -2683,15 +2687,386 @@ fn test_should_fail_with_zero_rewards() {
...
+        )
+        .unwrap();
+    assert_eq!(agent_reward, Some(Uint128::zero()));
+
+    // Agent contract can call WithdrawAgentRewards even if the reward is zero
+    let payable_account_balance_before_withdraw = app
+        .wrap()
+        .query_balance(PARTICIPANT2, DENOM)
+        .unwrap()
+        .amount;
```
Buckram123, line 3019: Is it right thing to do tho?

Buckram123 (approved):
Nice, thanks for testing it. Not sure if we should return error on empty agent-withdraw tho

PR #309, opened 2023-02-02
negative tests for tasks

Buckram123:
Few fixes: 

- reject empty actions tasks
- allow attaching funds to the wasm actions. We want to support it right @TrevorJTClarke @mikedotexe ?
  - Note: we were ignoring funds on the wasm action which is a vulnerability where we did let anyone attach any funds from manager balance to his tasks LOL :rofl:. From the `main` branch: https://github.com/CronCats/cw-croncat/blob/b920d1a5d5b1730397b04acdbba205eacfd487bd/packages/cw-croncat-core/src/types.rs#L264-L268


mikedotexe:
Good question @Buckram123 I think we need to definitely support attaching funds.
We don't want the CronCat contracts to be a place where people custody their funds, but there are certainly great use cases where you want to attach some nominal amount of funds.

For example, I want to wait until IBC Domains allows 3-letter domains like `cat.cosmos` and I'm willing to have CronCat hold onto 40 JUNO (or whatever) to pay for the registration when it's available. So again, at this point in time, we're not interested in people using CronCat as a vault or custody solution, but there will lots of value in allowing funds to be attached to tasks.

TrevorJTClarke:
> Good question @Buckram123 I think we need to definitely support attaching funds. We don't want the CronCat contracts to be a place where people custody their funds, but there are certainly great use cases where you want to attach some nominal amount of funds.
> 
> For example, I want to wait until IBC Domains allows 3-letter domains like `cat.cosmos` and I'm willing to have CronCat hold onto 40 JUNO (or whatever) to pay for the registration when it's available. So again, at this point in time, we're not interested in people using CronCat as a vault or custody solution, but there will lots of value in allowing funds to be attached to tasks.

im aligned on this as well

PR #310, opened 2023-02-02
[SPLIT] Update agents stats on task completion

deveusss:
This PR is for sending messages to agents contract on task completion


contracts/croncat-agents/src/contract.rs
```diff
@@ -689,3 +691,34 @@ fn on_task_created(
...
+
+    if args.is_block_slot_task {
+        stats.completed_block_tasks += 1;
+    } else {
+        stats.completed_cron_tasks += 1;
+    }
+    AGENT_STATS.save(deps.storage, &args.agent_id, &stats)?;
+
+    let response = Response::new()
+        .add_attribute("method", "on_task_completed")
```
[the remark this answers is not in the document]
deveusss: good catch, thx

contracts/croncat-agents/src/contract.rs
```diff
@@ -689,3 +691,34 @@ fn on_task_created(
...
+        .may_load(deps.storage, &args.agent_id)?
+        .ok_or(ContractError::AgentNotRegistered {})?;
+    croncat_manager_contract::assert_caller_is_manager_contract(
+        &deps.querier,
+        &config,
+        &info.sender,
+    )?;
+    let mut stats = AGENT_STATS
+        .may_load(deps.storage, &args.agent_id)?
+        .unwrap_or_default();
```
Buckram123, line 710:
```suggestion
    let mut stats = AGENT_STATS
        .load(deps.storage, &args.agent_id)?;
```
I don't like it being default here. It should just load, since we checked this agent is active already on `proxy_call`, and, in case, if it didn't load we better fail the transaction, instead of giving false agent stats progression!

contracts/croncat-manager/src/contract.rs
```diff
@@ -528,8 +530,21 @@ pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractE
...
+                let config = CONFIG.load(deps.storage)?;
+                //todo: fix is_block_slot param after boundary fix
+                let complete_msg = create_task_completed_msg(
+                    &deps.querier,
+                    &config,
+                    queue_item.task.task_hash.clone(),
+                    &queue_item.agent_addr,
+                    matches!(
+                        queue_item.task.interval,
+                        Interval::Block(_) | Interval::Immediate
```
Buckram123, line 542:
```suggestion
                        Interval::Block(_) | Interval::Immediate | Interval::Once
```
Once is also always should be a cron

packages/croncat-sdk-core/src/internal_messages/agents.rs
```diff
@@ -1,11 +1,18 @@
...
+use cosmwasm_std::{to_binary, Addr, Binary, CosmosMsg, StdResult, WasmMsg};
 
 #[cw_serde]
 pub struct AgentOnTaskCreated {
     pub task_hash: String,
 }
 
+#[cw_serde]
+pub struct AgentOnTaskCompleted {
+    pub task_hash: String,
```
Buckram123, line 11:
```suggestion
```
Let's have minimal amount of data transfer, can't see it's needed right now

Buckram123 (changes requested):
Nice, it's so close!

contracts/croncat-manager/src/contract.rs
deveusss: We need block interval here @Buckram123

Buckram123: nvm it could be just
 ```
!matches!(
                        queue_item.task.interval,
                        Interval::Cron(_)
                        ),
  ```

mikedotexe (approved):

PR #312, opened 2023-02-03
E2E MAXX - generating all scenarios of tasks to prove coverage works

TrevorJTClarke:
adding more helper methods for generating tasks


scripts/deployment/src/taskSampleData.ts
```diff
@@ -0,0 +1,469 @@
...
+  Object.keys(queries).forEach(k => {
+    Object.keys(queries[k]).forEach(d => {
+      keys.push(d)
+    })
+  })
+
+  return keys
+}
+
+// The totes magix AMIRIGHT
```
mikedotexe, line 296: You are correct, yes 🧙🏼

Issue #314, opened 2023-02-03
TODOs from the code comments


Buckram123:
- [ ] After juno stable release 1.2 arrives - update factory to use Intantiate2 for deployments. 
- [ ] Make IndexedMap consts, as soon as new version of `cw-storage-plus` arrives. Current version: "1.0.1". 
- [ ] Try out how we can migrate our contracts from the factory contract

PR #315, opened 2023-02-03
Todos cleanup

Buckram123:
Moved semi-blocked items and migrate(which is not priority for now) from the code to the #314 Issue


packages/croncat-sdk-manager/src/types.rs
```diff
@@ -192,7 +192,7 @@ pub struct Config {
 
     // Treasury
     pub treasury_addr: Option<Addr>,
-    pub cw20_whitelist: Vec<Addr>, // TODO: Consider fee structure for whitelisted CW20s
+    pub cw20_whitelist: Vec<Addr>,
```
Buckram123, line 195: Very important: Do we want to support any cw20s or only whitelisted? @TrevorJTClarke

Buckram123: Not using cw20 whitelist can give users ability to lock contract. 
Few ways to do so: 

- if someone creates a wrong cw20 implementation to intentionally fail send/transfer when croncat-manager is sender. (Could be covered, if we whitelist code_ids)
- If someone creates a cw20 action then migrates to the defective contract to change the state of the croncat-balance to the zero and migrate back to the supported code_id (Couldn't get covered with whitelisting code_ids)

So I'll make sure we allow only supported cw20 addresses

contracts/croncat-factory/src/contract.rs
```diff
@@ -48,7 +48,6 @@ fn init_save_metadata_generate_wasm_msg(
     )?;
     LATEST_VERSIONS.save(storage, &init_info.contract_name, &init_info.version)?;
 
-    // TODO: Once cosmwasm 1.2 is available, change to Instantiate2 https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/results/cosmos_msg.rs#L169
```
TrevorJTClarke, line 51: Hey - i still want this here :D maybe change to `TODO-FUTURE:`

contracts/croncat-tasks/src/state.rs
```diff
@@ -23,7 +23,6 @@ pub const TIME_MAP_QUERIES: Map<&[u8], u64> = Map::new("time_map_queries");
 /// Block based map by the corresponding task hash
 pub const BLOCK_MAP_QUERIES: Map<&[u8], u64> = Map::new("block_map_queries");
 
-// TODO: make IndexedMap's const as soon as cw_storage_plus new version arrives
```
TrevorJTClarke, line 26: @Buckram123 can you give me TLDR?

packages/croncat-sdk-manager/src/types.rs
TrevorJTClarke: nice, i def like the strategy here. We can always open up support later, better to restrict up front

TrevorJTClarke (approved):
@Buckram123 still have a few comments/qs - but everything here is a great update!!!

contracts/croncat-tasks/src/state.rs
Buckram123: Hope they release it soon, so we can make it consts. like other storage items:
 https://github.com/CosmWasm/cw-storage-plus/blob/main/CHANGELOG.md#unreleased

PR #318, opened 2023-02-03
Small change to simplify bank sends

Buckram123:
Since empty sends don't actually fail tx we can simplify it


TrevorJTClarke (approved):
nice!

PR #324, opened 2023-02-03
attribute vaccuum session

TrevorJTClarke:
### Rationale

for response attributes, i've cleaned up to optimize for gas with the following criteria to keep:
* Only if we don't already know the data from input (Example: Agent register, we know sender is the agent, dont need attribute)
* Annotate action, such that indexer can follow the entire flow completed, or potentially flag issues if didn't exist
* Additional information difficult to query or compute


contracts/croncat-manager/src/contract.rs
```diff
@@ -306,7 +306,7 @@ fn execute_proxy_call_with_queries(
         };
         Ok(Response::new()
             .add_attribute("action", "remove_task")
-            .add_attribute("task_status", "invalid")
+            .add_attribute("lifecycle", "task_invalidated")
```
mikedotexe, line 309: hmm that's cool, i dig it

mikedotexe (approved):

contracts/croncat-manager/src/contract.rs
TrevorJTClarke: tick also has the same naming - i was thinking future selves can alert on lifecycle things

Buckram123 (approved):
Really like those changes!

PR #326, opened 2023-02-04
Boundary fix

Buckram123:
TODO: 

- [x]  fix tests ~~(haha now it fails, cause we had `Interval::Once` into `Boundaryk:Time`: and etc.)~~
- [x] add new tests


contracts/croncat-tasks/src/helpers.rs
```diff
@@ -17,43 +17,48 @@ use crate::{
...
+    boundary: Option<Boundary>,
     interval: &Interval,
-) -> Result<BoundaryValidated, ContractError> {
-    let boundary_validated = match (interval, boundary) {
-        (Interval::Cron(_), Some(Boundary::Height { .. }))
-        | (Interval::Block(_), Some(Boundary::Time { .. })) => {
-            Err(ContractError::InvalidBoundary {})
+) -> Result<Boundary, ContractError> {
+    match (interval, boundary) {
+        (Interval::Cron(_), Some(Boundary::Time(boundary_time))) => {
```
TrevorJTClarke, line 24: @Buckram123 - kinda had the opposite thought on this, can you validate my thinking and potentially adjust if makes sense?
I'm imagining someone creating a task with boundary start & end, and the task taking a while to get on-chain.
I request current block, start boundary is then the next block, but it gets into the chain a block after my start. 
Now my task creation fails. ughhhh

I actually am thinking end is the only thing needing validation here, so it cannot end before current block (or really end + interval). Starting in the past is fine, because we're really just doing modulo on top of it and reschedule doesnt use start as the last time task occurred. If we change to end, does that mess up logic in other places? I believe this is a better approach for UX.
Thoughts?

contracts/croncat-tasks/src/helpers.rs
```diff
@@ -310,3 +315,358 @@ pub(crate) fn task_with_queries_ready(
...
+                Interval::Cron("* * * * * *".to_owned()),
+                Some(Boundary::Time(BoundaryTime {
+                    start: Some(Timestamp::from_nanos(123456)),
+                    end: Some(Timestamp::from_nanos(123456)),
+                })),
+                123,
+                Timestamp::from_nanos(123456),
+                Err(ContractError::InvalidBoundary {}),
+            ),
+        ];
```
TrevorJTClarke, line 661: really appreciate the thoroughness here!

TrevorJTClarke (changes requested):
hey @Buckram123 this PR lookin super solid - check my thoughts below

contracts/croncat-tasks/src/helpers.rs
Buckram123: > I actually am thinking end is the only thing needing validation here, so it cannot end before current block (or really end + interval). Starting in the past is fine, because we're really just doing modulo on top of it and reschedule doesnt use start as the last time task occurred. If we change to end, does that mess up logic in other places? I believe this is a better approach for UX. Thoughts?

Thanks, strongly agree!

Buckram123: @TrevorJTClarke fixed!

TrevorJTClarke (approved):
Thank you!!!

PR #327, opened 2023-02-04
Event based tasks simplified storage update pro max plus plus grande 

TrevorJTClarke:
PR Summary
* Changes tasks to support queries for both scheduled & evented
* added a manual lookup map for tracking evented tasks
* updated all task & manager methods handling event/schedule logic


contracts/croncat-manager/src/contract.rs
```diff
@@ -245,9 +245,10 @@ fn execute_proxy_call_with_queries(
 
     // Get a task
     let tasks_addr = get_tasks_addr(&deps.querier, &config)?;
+    // TODO: Change to only check if its ready, get regular task by hash
     let current_task: croncat_sdk_tasks::types::TaskResponse = deps.querier.query_wasm_smart(
         tasks_addr.clone(),
-        &croncat_sdk_tasks::msg::TasksQueryMsg::CurrentTaskWithQueries {
+        &croncat_sdk_tasks::msg::TasksQueryMsg::Task {
```
Buckram123, line 251: So we can't schedule task with queries?

Buckram123 (commented):
So if this task is "event based" we can't reschedule it, right? What intervals/boundaries usable with it, shouldn't it be only `Once` interval?

contracts/croncat-manager/src/contract.rs
TrevorJTClarke: we totally do schedule tasks with queries - they're just not treated any diff than tasks without queries.
Still reworking the impl, so will need eyes to make sure I didn't miss anything. But the point will be to just simply do the queries at proxy_call, and not have a separate schedule/bucket for tasks with queries

TrevorJTClarke:
> What intervals/boundaries usable with it, shouldn't it be only `Once` interval?

@Buckram123 - intervals Once, Immediate still in play, as it helps us remove the task (or not in case of Immediate) based on funds. The boundary then is used for grouping which non-scheduled tasks are active for the agents

Buckram123:
> @Buckram123 - intervals Once, Immediate still in play, as it helps us remove the task (or not in case of Immediate) based on funds. The boundary then is used for grouping which non-scheduled tasks are active for the agents

Oh, right, supporting immediate makes sense, thanks!

contracts/croncat-manager/src/contract.rs
```diff
@@ -128,10 +128,8 @@ pub fn execute(
 ) -> Result<Response, ContractError> {
     match msg {
         ExecuteMsg::UpdateConfig(msg) => execute_update_config(deps, info, *msg),
-        ExecuteMsg::ProxyCall { task_hash: None } => execute_proxy_call(deps, env, info),
-        ExecuteMsg::ProxyCall {
-            task_hash: Some(task_hash),
-        } => execute_proxy_call_with_queries(deps, env, info, task_hash),
+        // ExecuteMsg::ProxyCall { task_hash: None } => execute_proxy_call(deps, env, info),
```
Buckram123, line 131:
```suggestion
```
I guess not needed anymore

contracts/croncat-manager/src/contract.rs
```diff
@@ -175,33 +173,118 @@ fn execute_remove_task(
...
+        &croncat_sdk_agents::msg::QueryMsg::GetAgent {
             account_id: info.sender.to_string(),
         },
     )?;
-    if agent_tasks.stats.num_block_tasks.is_zero() && agent_tasks.stats.num_cron_tasks.is_zero() {
+    if agent_reponse.agent.map_or(true, |agent| {
+        agent.status != croncat_sdk_agents::types::AgentStatus::Active
+    }) {
         return Err(ContractError::NoTaskForAgent {});
     }
```
Buckram123, line 195: That feels wrong, we don't want 1 agent to take all tasks? I guess it still should check `AgentTaskResponse` if no `task_hash` and `AgentResponse` if there is

contracts/croncat-manager/src/contract.rs
```diff
@@ -175,33 +173,118 @@ fn execute_remove_task(
...
+        // End task
+        return end_task(
+            deps,
+            task,
+            config,
+            tasks_addr,
+            task_balance,
+            Some(vec![Attribute::new("lifecycle", "task_ended")]),
+        );
+    }
```
Buckram123, line 234: Nice, I like that addition

contracts/croncat-manager/src/contract.rs
```diff
@@ -211,117 +294,39 @@ fn execute_proxy_call(
...
-        Ok(Response::new()
-            .add_attribute("action", "proxy_call_with_queries")
-            .add_submessages(sub_msgs))
-    }
+    Ok(Response::new()
+        .add_attribute("action", "remove_task")
+        .add_attributes(attrs.unwrap_or_default())
+        .add_message(msg)
+        .add_message(bank_send))
 }
```
Buckram123, line 330: Would like to reward agent for doing maiden's job

Buckram123 (approved):
Awesome!

contracts/croncat-manager/src/contract.rs
mikedotexe: Ultimately, this is checking to see if the sender is an active agent, which seems fine. It does give me pause, though, to think about this now.

I do miss the certainty we had before a bit. Like, the agent checks to see if they've got tasks to execute by calling `GetAgentTasks`, and then do a `proxy_call` and the contract them immediately performs `GetAgentTasks` to make sure it's reasonable to move forward.

Actually wait, I think I now understand what Misha said. One agent could just ignore asking, "how many should I do?" and just spam `proxy_call`. If they're in the active set, they'd skip through this check now

contracts/croncat-manager/src/contract.rs
```diff
@@ -175,33 +173,118 @@ fn execute_remove_task(
...
+        )?
+    } else {
+        // get a scheduled task
+        deps.querier.query_wasm_smart(
+            tasks_addr.clone(),
+            &croncat_sdk_tasks::msg::TasksQueryMsg::CurrentTask {},
+        )?
+    };
+
+    let Some(mut task) = current_task.task else {
```
mikedotexe, line 213: new syntax usage 👏🏼

contracts/croncat-manager/src/contract.rs
TrevorJTClarke: thats a good catch @Buckram123 -- i need switch based on if task_hash or not

contracts/croncat-manager/src/contract.rs
Buckram123: really enjoy `let else` thingy

contracts/croncat-tasks/src/helpers.rs
```diff
@@ -190,14 +187,34 @@ pub(crate) fn validate_msg_calculate_usage(
...
+            .collect::<StdResult<Vec<_>>>()?;
+        for (hid, mut all_hashes) in hashes {
+            let mut found = false;
+            all_hashes.retain(|h| {
+                found = h == hash;
+                !found
+            });
+            if found {
+                if all_hashes.is_empty() {
+                    BLOCK_SLOTS.remove(storage, hid);
```
mikedotexe, line 210: are we sure it's in the `BLOCK_SLOTS` Map and not `TIME_SLOTS`?

TrevorJTClarke: oh snap - nono, you found that i made bad copypasta: should be `EVENTED_TASKS_LOOKUP`

contracts/croncat-tasks/src/helpers.rs
```diff
@@ -296,23 +298,6 @@ pub(crate) fn get_agents_addr(
         .ok_or(ContractError::InvalidKey {})
 }
 
-/// Check that this task can be executed in current slot
-pub(crate) fn task_with_queries_ready(
```
mikedotexe, line 300: Surprised this is no longer needed? Or was it simply unused?

TrevorJTClarke: this section replaced the need for that: https://github.com/CronCats/cw-croncat/blob/ft/task-query-consolidate/contracts/croncat-manager/src/contract.rs#L220-L240

contracts/croncat-manager/src/contract.rs
TrevorJTClarke: good call

contracts/croncat-manager/src/contract.rs
TrevorJTClarke: updated

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -365,13 +374,35 @@ fn get_next_block_limited(env: &Env, boundary_height: &BoundaryHeight) -> (u64,
...
+    let current_block_time = env.block.time.nanos();
+
+    let next_block_time = match boundary.start {
+        Some(id) if current_block_time < id.nanos() => id.nanos(),
+        _ => current_block_time,
+    };
+
+    match boundary.end {
+        // stop if passed end time
+        Some(end) if current_block_time > end.nanos() => 0,
```
mikedotexe, line 399: Doesn't this return the timestamp of `0` (so Linux epoch in the 70s), and then that's used in the Interval's impl for `next`, meaning we report a task with a timestamp in the past, which generally means the agent should handle it as it does it other, earlier tasks?

I'm trying to follow the logic…

mikedotexe: Wait a second, it's returning the ID, not the timestamp in nanos as the first value in the tuple, then

```rs
    if next_id == 0 {
        return Err(ContractError::TaskEnded {});
    }
```

nvm. I'll add comments above that impl for help with return types

TrevorJTClarke: yeah - sorry its a little convoluted because the use is in a far diff spot: https://github.com/CronCats/cw-croncat/blob/ft/task-query-consolidate/contracts/croncat-tasks/src/contract.rs#L169

Essentially if `next_id == 0` ever, we know to not reschedule

mikedotexe (approved):

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -99,22 +99,31 @@ impl Interval {
             // If Once, return the first block within a specific range that can be triggered 1 time.
             // If Immediate, return the first block within a specific range that can be triggered immediately, potentially multiple times.
             (Interval::Once, Boundary::Height(boundary_height))
-            | (Interval::Immediate, Boundary::Height(boundary_height)) => {
-                get_next_block_limited(env, boundary_height)
+            | (Interval::Immediate, Boundary::Height(boundary_height)) => (
+                get_next_block_limited(env, boundary_height),
+                SlotType::Block,
+            ),
```
Buckram123, line 105: Nice, cool!

Buckram123 (approved):
Masterpiece

contracts/croncat-manager/src/contract.rs
```diff
@@ -175,153 +172,199 @@ fn execute_remove_task(
...
-    let agent_tasks: croncat_sdk_agents::msg::AgentTaskResponse = deps.querier.query_wasm_smart(
-        agents_addr,
-        &croncat_sdk_agents::msg::QueryMsg::GetAgentTasks {
-            account_id: info.sender.to_string(),
-        },
-    )?;
-    if agent_tasks.stats.num_block_tasks.is_zero() && agent_tasks.stats.num_cron_tasks.is_zero() {
-        return Err(ContractError::NoTaskForAgent {});
-    }
+    if task_hash.is_none() {
```
Buckram123, line 185: litl nit, note that you checked twice task_hash option, first here

contracts/croncat-manager/src/contract.rs
```diff
@@ -175,153 +172,199 @@ fn execute_remove_task(
...
+    // Get a task
     let tasks_addr = get_tasks_addr(&deps.querier, &config)?;
-    let current_task: croncat_sdk_tasks::types::TaskResponse = deps.querier.query_wasm_smart(
-        tasks_addr,
-        &croncat_sdk_tasks::msg::TasksQueryMsg::CurrentTask {},
-    )?;
-    // Note: there should be at least one task
-    // unless `get_agent_tasks` gave false positive
-    let task = current_task.task.unwrap();
+    let current_task: croncat_sdk_tasks::types::TaskResponse = if let Some(hash) = task_hash {
```
Buckram123, line 215: second here

contracts/croncat-manager/src/contract.rs
```diff
@@ -175,153 +172,199 @@ fn execute_remove_task(
...
+            config,
+            agent_addr,
+            tasks_addr,
+            Some(vec![Attribute::new("lifecycle", "task_ended")]),
+        );
+    }
+
+    if let Some(queries) = task.queries.as_ref() {
+        let event_based = queries.iter().any(|q| q.check_result);
+        if event_based && !(task.interval == Interval::Once || task.interval == Interval::Immediate)
```
Buckram123, line 252: is there any case it can get created with `Block` or `Immediate` interval? We can't change it after creation. Does this check make sense?

Buckram123 (approved):
littliest nits

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -247,8 +256,8 @@ impl Task {
         !matches!(self.interval, Interval::Once)
     }
 
-    pub fn with_queries(&self) -> bool {
-        !self.queries.is_empty()
+    pub fn is_evented(&self) -> bool {
+        self.queries.iter().any(|q| q.check_result)
     }
```
Buckram123, line 261: Maybe add `&& self.interval == Interval::Once || (self.interval == Interval::Immediate)` ?

contracts/croncat-manager/src/contract.rs
mikedotexe: I was just looking here, too.

I wrote a test for a task that has a query with `check_result = true`. I'm having the query return `false`, and getting back `NoTaskForAgent` which seemed odd. I was thinking it would be `TaskNotReady`

mikedotexe: (I'm not positive it's being returned here, btw, but looking now 👀 )

TrevorJTClarke: Shouldn't be - i just figured better to safe than sorry :P

packages/croncat-sdk-tasks/src/types.rs
TrevorJTClarke: i could - but i wanted the simplest path to check, the interval should be validated separately i think

TrevorJTClarke: added update

packages/croncat-sdk-tasks/src/msg.rs
```diff
@@ -75,21 +75,36 @@ pub enum TasksQueryMsg {
...
+        limit: Option<u64>,
+    },
+    #[returns(Vec<String>)]
+    EventedHashes {
+        id: Option<u64>,
+        from_index: Option<u64>,
+        limit: Option<u64>,
+    },
+    /// Get list of event driven tasks
     #[returns(Vec<crate::types::TaskResponse>)]
```
mikedotexe, line 105: This doesn't match the return value here, I don't think:

https://github.com/CronCats/cw-croncat/pull/327/files#diff-d773b8fd5f2b6ed20164da89f72a626db0e8400517f0365e3fe46309999763f0R618-R624

Buckram123: Is there any tool that can check those? If not - I'm starting that project tomorrow, should be really useful

Buckram123: Not everyone has @mikedotexe eyes in the pockets

mikedotexe, line 105:
```suggestion
    #[returns(Vec<crate::types::TaskInfo>)]
```

TrevorJTClarke: thanks - fixed

contracts/croncat-tasks/src/contract.rs
```diff
@@ -353,6 +350,14 @@ fn execute_create_task(
         transforms: task.transforms.unwrap_or_default(),
         version: config.version.clone(),
     };
+    let event_based = item.is_evented();
+    if !item.interval.is_valid()
+        || (event_based
+            && !(item.interval == Interval::Once || item.interval == Interval::Immediate))
```
Buckram123, line 356: Nice!

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -258,6 +258,7 @@ impl Task {
 
     pub fn is_evented(&self) -> bool {
         self.queries.iter().any(|q| q.check_result)
+            && (self.interval == Interval::Once || self.interval == Interval::Immediate)
```
Buckram123, line 261: Still needed?

contracts/croncat-manager/src/contract.rs
```diff
@@ -175,153 +172,198 @@ fn execute_remove_task(
...
-    )?;
-    if agent_tasks.stats.num_block_tasks.is_zero() && agent_tasks.stats.num_cron_tasks.is_zero() {
-        return Err(ContractError::NoTaskForAgent {});
+    let tasks_addr = get_tasks_addr(&deps.querier, &config)?;
+
+    // Check if agent is active,
+    // Then get a task
+    let current_task: croncat_sdk_tasks::types::TaskResponse = if let Some(hash) = task_hash {
+        // For evented case - check the agent is active, then may the best agent win
+        let agent_reponse: croncat_sdk_agents::msg::AgentResponse = deps.querier.query_wasm_smart(
```
mikedotexe, line 190:
```suggestion
        let agent_response: croncat_sdk_agents::msg::AgentResponse = deps.querier.query_wasm_smart(
```

contracts/croncat-manager/src/contract.rs
```diff
@@ -175,153 +172,198 @@ fn execute_remove_task(
...
+            agents_addr,
+            &croncat_sdk_agents::msg::QueryMsg::GetAgent {
+                account_id: agent_addr.to_string(),
+            },
+        )?;
+        if agent_reponse.agent.map_or(true, |agent| {
```
mikedotexe, line 196:
```suggestion
        if agent_response.agent.map_or(true, |agent| {
```

packages/croncat-sdk-tasks/src/types.rs
Buckram123: It's really wrong now, evented task with block/cron will just slide into schedule. So it's better to remove this, right?

TrevorJTClarke: sure - will remove

Buckram123 (approved):
Awesome!

PR #331, opened 2023-02-05
For packages, add necessary manifest fields

mikedotexe:
All the packages are on crates.io now, and in that process it makes us add useful things like a description.

Did the thing

### DIFF
--- a/contracts/mod-generic/Cargo.toml
+++ b/contracts/mod-generic/Cargo.toml
@@ -5,7 +5,7 @@ authors = []
 edition = "2021"
 repository = "https://github.com/CronCats/cw-croncat"
 license = { workspace = true }
-description = "Allows for CosmWasm raw queries through this module, helpful when making CronCat tasks with queries"
+description = "Allows to deserialize any CosmWasm smart contract's queue response and check the results of any fields."
 
 [lib]
 crate-type = ["cdylib", "rlib"]
--- a/contracts/mod-generic/Cargo.toml
+++ b/contracts/mod-generic/Cargo.toml
@@ -4,6 +4,8 @@ version = "0.1.0"
 authors = []
 edition = "2021"
 repository = "https://github.com/CronCats/cw-croncat"
+license = { workspace = true }
+description = "Allows for CosmWasm raw queries through this module, helpful when making CronCat tasks with queries"
 
 [lib]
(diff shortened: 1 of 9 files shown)


contracts/mod-generic/Cargo.toml
```diff
@@ -4,6 +4,8 @@ version = "0.1.0"
 authors = []
 edition = "2021"
 repository = "https://github.com/CronCats/cw-croncat"
+license = { workspace = true }
+description = "Allows for CosmWasm raw queries through this module, helpful when making CronCat tasks with queries"
```
Buckram123, line 8: This contract is actually to make queries to any contract and check any field of it. I guess something like "Allows to deserialize any contract's queue response and check the results of any fields"

packages/croncat-sdk-manager/Cargo.toml
```diff
@@ -4,6 +4,8 @@ version = "0.1.0"
 authors = { workspace = true }
 edition = "2021"
 repository = "https://github.com/CronCats/cw-croncat"
+license = { workspace = true }
+description = "CronCat Manager types, messages, and errors."
```
Buckram123, line 8: I guess this one missing part about cosmwasm, but maybe I like this description even more

packages/croncat-sdk-tasks/Cargo.toml
```diff
@@ -4,6 +4,8 @@ version = "0.1.0"
 authors = { workspace = true }
 edition = "2021"
 repository = "https://github.com/CronCats/cw-croncat"
+license = { workspace = true }
+description = "CronCat tasks package containing CosmWasm types and messages"
```
Buckram123, line 8: Doesn't it have errors?

Buckram123 (approved):
Thanks! Don't know if we have to make standardized descriptions

contracts/mod-generic/Cargo.toml
mikedotexe: sure, I'll change it to that

packages/croncat-sdk-tasks/Cargo.toml
mikedotexe: Not yet 😛 

Screen Shot 2023-02-04 at 7 28 50 PM

PR #332, opened 2023-02-05
Adding tests to ft/task-query-consolidate

mikedotexe:
Tests we discussed:

- [x] ✅  event-type task is not executed by `proxy_call` when end boundary is surpassed (`event_task_with_boundary_issue`)
- [x] ✅  regular task is not executed by `proxy_call` when end boundary is surpassed (`scheduled_task_with_boundary_issue`)
- [x] ✅ event-type task queries evaluate to false and task isn't fulfilled, remaining ready
- [x] ✅ event-type task with Immediate interval executes more than once when criteria is met
- [x] ✅ event-type task with Once interval cannot be executed multiple times


mikedotexe:
I don't know why the CI's checksum is acting up, but I'm not terribly worried about that. Our plan has been to have CI create it for us. I'm inclined to merge without that passing personally

TrevorJTClarke (approved):
This is some great coverage! thank you!!!

PR #335, opened 2023-02-06
README in query contracts and packages

NinoLipartiia:
There are new READMEs for query contracts, similar to docs we have for agent, manager and tasks contracts.

Also added small READMEs for packages, let me know if these should be more detailed or include the same tables as in the docs for corresponding contracts.


deveusss:
thank you, @NinoLipartiia 👍🏻

PR #339, opened 2023-02-08
Replace vectors of task hashes

### LINKED ISSUE #272: Stabilize time/block slots, so it won't lock our contract or increase gas during proxy call
https://github.com/CronCats/cw-croncat/blob/b920d1a5d5b1730397b04acdbba205eacfd487bd/contracts/cw-croncat/src/state.rs#L117-L121

Value can contain infinite amount of element and we can't predict how much gas is gonna be used to reward agents.

Idea from @deveusss is to get rid of those two maps and use `MultiIndex` inside task's `IndexedMap`

Buckram123:
Closes #272.  Do we need `query_slot_ids` ?

Doing the same to the evented tasks


Buckram123:
2 methods currently broken, but I don't think we need them, except for the debugging: `query_slot_ids` and `query_evented_ids`. Making slots more complex, just for debugging doesn't look like a worthy thing, thoughts @TrevorJTClarke @mikedotexe?

contracts/croncat-tasks/src/contract.rs
```diff
@@ -631,48 +566,42 @@ fn query_evented_tasks(
...
+            .map(|res| res.map(|v| v.0))
+            .collect::<StdResult<_>>()?
     } else {
-        let block_evented: Vec<(u64, _)> = EVENTED_TASKS_LOOKUP
-            .range(
+        let block_evented_iter = EVENTED_TASKS_LOOKUP
+            .prefix_range(
                 deps.storage,
-                Some(Bound::inclusive(env.block.height)),
+                // TODO: should this boundary be min or max
```
Buckram123, line 582: @TrevorJTClarke It's min here, should it be max instead?

Buckram123: Or actually: why boundary here at all?

deveusss:
hey @Buckram123 , good job

TrevorJTClarke:
Still thinking through the ordering part of this - I really like most of the approach! 
Can you update the conflicts now @Buckram123 ?

PR #340, opened 2023-02-09
Beta/split 0.0.0 new agent distributor

deveusss:
Get rid of active agents vector and change distributor logic, so it respects agent activity


TrevorJTClarke:
can you update the conflicts @deveusss?

PR #346, opened 2023-02-15
queries-transforms-fallback

Buckram123:
- Fixes a bug, where we lock our tasks queue due to the failed query or transform.
- Fixes a bug where we did allow to execute any task by `task_hash`


TrevorJTClarke:
can you update the conflicts @Buckram123?

PR #349, opened 2023-02-22
Fix self call uppercase issue

TrevorJTClarke:


mikedotexe (approved):
awesome, look forward to 350

PR #351, opened 2023-02-23
Item 6: Config verification for agents, test fixes

mikedotexe:
This mostly is changing how the agent tests are working. Previously, it was assuming that an admin would call it, but we want to make sure our tests are always calling the Factory, and then using the `proxy` function to update config and whatnot.

Also added checks in an agent method


contracts/croncat-agents/src/contract.rs
```diff
@@ -680,3 +695,63 @@ fn on_task_completed(
     let response = Response::new().add_attribute("action", "on_task_completed");
     Ok(response)
 }
+
+/// Ensure min_tasks_per_agent does not accept zero during
+/// instantiate or update config assignments
+fn validate_min_tasks_per_agent(val: Option<u64>) -> Result<(), ContractError> {
```
TrevorJTClarke, line 701: Can these fns be changed to 1 function? They're all validating the same thing, just using diff data

contracts/croncat-agents/src/error.rs
```diff
@@ -56,4 +56,7 @@ pub enum ContractError {
 
     #[error("No rewards available for withdraw")]
     NoWithdrawRewardsAvailable {},
+
+    #[error("No rewards available for withdraw")]
```
TrevorJTClarke, line 60: Can this change? Looks like copy from above, and doesnt really make sense with the enum value below

contracts/croncat-agents/src/tests/agent_tests.rs
```diff
@@ -139,13 +141,20 @@ fn test_register_agent_fails() {
...
+    // Check contract is paused and failing
     let mut config = mock_update_config(croncat_factory_addr.as_str());
     config.paused = Some(true);
+
     app.execute_contract(
         Addr::unchecked(ADMIN),
-        croncat_agents_addr.clone(),
-        &ExecuteMsg::UpdateConfig { config },
+        croncat_factory_addr,
+        &croncat_sdk_factory::msg::FactoryExecuteMsg::Proxy {
```
TrevorJTClarke, line 151: amazing, thank you!

contracts/croncat-agents/src/tests/agent_tests.rs
```diff
@@ -249,10 +258,17 @@ fn test_update_agent_fails() {
     //Check contract is paused and failing
     let mut config = mock_update_config(croncat_factory_addr.as_str());
     config.paused = Some(true);
+
     app.execute_contract(
         Addr::unchecked(ADMIN),
-        croncat_agents_addr.clone(),
-        &ExecuteMsg::UpdateConfig { config },
```
TrevorJTClarke, line 255: I guess it would make sense to ensure a fail case happens from non-factory calling this

contracts/croncat-agents/src/tests/agent_tests.rs
```diff
@@ -1040,25 +1088,45 @@ fn test_tick() {
         )
         .unwrap();
     assert_eq!(agents.active.len(), 1);
-    assert_eq!(agents.pending.len(), 2);
+    assert_eq!(agents.pending.len(), 3);
+
+    // trying stuff before meeting, can remove
```
TrevorJTClarke, line 1093: following up here - did you intend to leave?

TrevorJTClarke (changes requested):
Awesome updates esp to test cov! The proxy via factory is soooooo good @mikedotexe -- left a couple changes below

contracts/croncat-agents/src/error.rs
mikedotexe: Ha, yes, I saw this right before bed but didn't fix it yet. Fixed in https://github.com/CronCats/cw-croncat/pull/351/commits/b56f70e36d9e8030bb2a64a68a4decc8acc70359

contracts/croncat-agents/src/tests/agent_tests.rs
mikedotexe: good eye, removed here: https://github.com/CronCats/cw-croncat/pull/351/commits/4c40302e3a21ec12a122d40e12e4d1567708b775

contracts/croncat-agents/src/tests/agent_tests.rs
mikedotexe: Extra failure cases added here: https://github.com/CronCats/cw-croncat/pull/351/commits/3bfcc4356ca3f1b1fc8e1f763d407faf07b67b28

contracts/croncat-agents/src/contract.rs
TrevorJTClarke: interesting! didn't know

contracts/croncat-agents/src/tests/agent_tests.rs
```diff
@@ -255,10 +255,55 @@ fn test_update_agent_fails() {
...
+        .downcast()
+        .unwrap();
+    assert_eq!(
+        factory_err,
+        croncat_factory::ContractError::Unauthorized {},
+        "Only factory admin can update config"
+    );
+
+    // Direct call to update config should also fail
+    let mut agents_err: ContractError = app
```
TrevorJTClarke, line 289: Eggsellent!

TrevorJTClarke (approved):
TYSM!

PR #358, opened 2023-03-01
Item 6: Manager Config value validation

mikedotexe:
Here we're validating the fee and treasury fees in config are between 0 - 100.

I'll note that while I think it's a good idea to address this, it's possible that chains with relatively low native token prices might actually work better if we're allowed to have greater than 100%, but we can have that conversation at another time.

I also saw that we're using `u64` for a number that's a maximum of 100, so I changed it to `u8` here.

Since this is on every `Task` object, I believe this'll save 14 bytes per task if my napkin math is right.


contracts/croncat-manager/src/contract.rs
```diff
@@ -644,10 +644,13 @@ pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractE
     }
 }
 
-/// Validate when a given value should be between 0 - 100.
+/// Validate when a given value should be a reasonable percentage.
+/// Due to low native token prices on some chains, we must allow for
+/// greater than 100% in order to be sustainable, and have gone with
+/// a max of 10,000% after internal discussion and looking at the numbers.
```
TrevorJTClarke, line 650: awesome!

TrevorJTClarke (approved):
TYSM!

PR #359, opened 2023-03-01
Item 6: Validate denom on Manager contract

mikedotexe:
We use the query for the bonded denom to validate whether the provided denom is valid.

There were a few cases across other contracts that used hardcoded values that then needed to be updated, as well.

UPDATE: removed the `denom` from the Manager's instantiate message as discussed in the comments here. Also modified the CosmJS deploy script accordingly.


contracts/croncat-manager/src/tests/manager_tests.rs
```diff
@@ -128,14 +125,14 @@ mod instantiate_tests {
...
-            native_denom: "cron".to_owned(),
+            native_denom: DENOM.to_string(),
             limit: 100,
             treasury_addr: Some(Addr::unchecked(AGENT2)),
         };
         assert_eq!(config, expected_config);
 
         let manager_balances = query_manager_balances(&app, &manager_addr);
-        assert_eq!(manager_balances, Uint128::new(600));
+        assert_eq!(manager_balances, Uint128::new(2400));
```
mikedotexe, line 135: Note that I took some time to make sure this wasn't just "changing the test to match the output" and indeed verified this tracks.

contracts/croncat-agents/src/tests/common.rs
```diff
@@ -30,7 +30,9 @@ pub const PARTICIPANT5: &str = "cosmos5t5u0jfg3ljsjrh2m9e47d4ny2hea7eehxrzdgd";
 pub const PARTICIPANT6: &str = "cosmos6t5u0jfg3ljsjrh2m9e47d4ny2hea7eehxrzdgd";
 pub const PARTICIPANT7: &str = "cosmos7t5u0jfg3ljsjrh2m9e47d4ny2hea7eehxrzdgd";
 
-pub const NATIVE_DENOM: &str = "uatom";
+/// We set this to "TOKEN" to match the denom here:
+/// https://github.com/CosmWasm/cosmwasm/blob/32f308a1a56ae5b8278947891306f7a374c3df94/packages/vm/src/environment.rs#L383
```
TrevorJTClarke, line 34: ah nice, was wondering about that!

contracts/croncat-manager/src/contract.rs
```diff
@@ -63,6 +63,16 @@ pub fn instantiate(
         cw20_whitelist,
     } = msg;
 
+    // Validate the denom, verify as case-insensitive, use bonded_denom value
+    let bonded_denom = deps.querier.query_bonded_denom()?;
+    let denom = if denom.to_uppercase() != bonded_denom.to_uppercase() {
```
TrevorJTClarke, line 68: @mikedotexe the goal is to actually remove instantiate denom, so we only rely on the protocol native denom

contracts/croncat-manager/src/tests/manager_tests.rs
TrevorJTClarke: I see the value above, fine w / me

contracts/croncat-manager/src/tests/manager_tests.rs
```diff
@@ -4698,3 +4695,70 @@ fn immediate_event_task_has_multiple_executions() {
...
+
+/// We test that the denom provided matches the validated, bonded denom value
+#[test]
+fn invalid_denom() {
+    let mut app = default_app();
+    let factory_addr = init_factory(&mut app);
+
+    let mut instantiate_msg: InstantiateMsg = default_instantiate_message();
+    // Test different, invalid denom
+    instantiate_msg.denom = "mochi".to_string();
```
TrevorJTClarke, line 4707: You did good work on this test - but as per above, i think we shouldn't use instantiate denom

TrevorJTClarke (changes requested):
Thanks for doing this!! See comments:

contracts/croncat-manager/src/contract.rs
mikedotexe: Ah, gotcha

TrevorJTClarke (approved):
Brilliant!

PR #363, opened 2023-03-02
Item 6: Tasks config validation

mikedotexe:
Validate gas_limit and slot_granularity_time during instantiate and config updates for the Task contract.


TrevorJTClarke (approved):
Nice!!

PR #365, opened 2023-03-02
21 limit coins

TrevorJTClarke:


contracts/croncat-manager/src/contract.rs
```diff
@@ -11,7 +11,7 @@ use croncat_sdk_manager::msg::AgentWithdrawCallback;
 use croncat_sdk_manager::types::{TaskBalance, TaskBalanceResponse, UpdateConfig};
 use croncat_sdk_tasks::types::{Interval, Task, TaskInfo};
 use cw2::set_contract_version;
-use cw_utils::parse_reply_execute_data;
+use cw_utils::{may_pay, parse_reply_execute_data};
```
mikedotexe, line 14: Nice find, I had not been familiar with `cw_utils` until now

mikedotexe (approved):

contracts/croncat-manager/src/contract.rs
TrevorJTClarke: i think its actually recent - i remember only seeing "has_coins" before

PR #367, opened 2023-03-03
8 initial setup for pause role on agents contract

TrevorJTClarke:
Key updates here:
1. Added pause role, which allows only a pause entity to pause the contract. Cannot unpause. Cannot be changed either, expected to be the pause entity for the duration of a version.
2. Owner was changed for agents, manager, tasks - to only ever be factory. This helps guard against strange side-attacks. For the duration of the version, the deploying factory will always be the owner.
3. The only way to unpause will be via the DAO that owns the factory. Which can help make deliberate just reasoning after a pause has occurred.


contracts/croncat-agents/src/contract.rs
```diff
@@ -526,14 +536,11 @@ pub fn execute_update_config(
         }
 
         let new_config = Config {
-            owner_addr: owner_addr
-                .map(|human| deps.api.addr_validate(&human))
-                .transpose()?
-                .unwrap_or(config.owner_addr),
+            owner_addr: config.owner_addr,
+            pause_admin: config.pause_admin,
```
mikedotexe, line 540: We cannot update the pause address?

mikedotexe: Ah, I see the description now, thx

mikedotexe (approved):

PR #371, opened 2023-03-03
reduce redundant code

TrevorJTClarke:
sometimes we can be blind to our own implementation 😹


mikedotexe (approved):
Kind of surprised we didn't get "unused method…" when doing all the linting, clippy, fmt for warnings, etc.

Thanks for the fix

PR #373, opened 2023-03-03
Added support for migrations within factory proxy

TrevorJTClarke:


packages/croncat-sdk-manager/src/types.rs
```diff
@@ -285,11 +285,6 @@ mod test {
             cw20_balance: Some(cw20.clone()),
             ibc_balance: Some(ibc_coin.clone()),
         };
-        // TODO:
```
mikedotexe, line 288: 🧹

mikedotexe (approved):
Future proof now!

PR #374, opened 2023-03-05
Item 4: Add whitelist for agent registration

mikedotexe:
As per some suggestions, we're introducing an agent registration whitelist. Progressive decentralization makes sense, and we'll use changes here.

The Agent contract's `Config` will have a new boolean field `public_registration` that defaults to `false`.

There's one more piece of state added: a `Map` that's being used like a set. It's called `APPROVED_AGENTS`.

A list of approved agents can be provided during instantiation or by calling the new method `add_agent_to_whitelist` which only supports/handles one insertion at a time. For removal, the function `remove_agent_from_whitelist` is introduced, also one at a time. Note that when the configuration updates include changing the `public_registration` from `false` to `true`, the whitelist will be cleared. This is intentionally a one-way boolean, used as machinery for a safe, controlled, progressive decentralization.

A new query method, `get_approved_agent_addresses` that uses pagination.

When a prospective agent calls `register_agent`, it fill fail with a new error unless the sender is on the approved list.


contracts/croncat-agents/src/tests/agent_tests.rs
```diff
@@ -1845,6 +1859,326 @@ fn pause_admin_cases() {
...
+    // Remove an agent
+    assert!(
+        app.execute_contract(
+            Addr::unchecked(ADMIN),
+            croncat_factory_addr.clone(),
+            &croncat_sdk_factory::msg::FactoryExecuteMsg::Proxy {
+                msg: WasmMsg::Execute {
+                    contract_addr: croncat_agents_addr.to_string(),
+                    msg: to_binary(&ExecuteMsg::RemoveAgentFromWhitelist {
+                        agent_address: AGENT0.to_string(),
```
TrevorJTClarke, line 2105: I missed commenting in the logic side, but it should be noted: Removing from whitelist doesnt remove agent from pending/active queue - and this is intentional. Just means that once they get removed from pending/active, thats it, they cannot register back in

contracts/croncat-agents/src/tests/agent_tests.rs
```diff
@@ -1845,6 +1859,326 @@ fn pause_admin_cases() {
...
+        .unwrap();
+    assert_eq!(
+        current_allowed_agents,
+        GetApprovedAgentAddresses {
+            approved_addresses: vec![Addr::unchecked(AGENT1)],
+        }
+    );
+
+    // Update config to allow public registration for agents
+    assert!(
```
TrevorJTClarke, line 2134: Based on above comment in contract, we'll need 1 more test to cover trying to set it back to False after true. Should fail

TrevorJTClarke (changes requested):
Looking pretty close! Thanks for the deep dive into this @mikedotexe -- see below

packages/croncat-sdk-agents/src/msg.rs
```diff
@@ -100,6 +117,12 @@ pub struct GetAgentIdsResponse {
     /// Pending agent list
     pub pending: Vec<Addr>,
 }
+/// Response containing approved agents' addresses
+#[cw_serde]
+pub struct GetApprovedAgentAddresses {
```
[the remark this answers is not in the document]
mikedotexe: good point, that is silly name for a response. fixed in https://github.com/CronCats/cw-croncat/pull/374/commits/677f58f373e6c51cbf0e110fe0e6d34fe34cd65e

contracts/croncat-agents/src/contract.rs
```diff
@@ -546,10 +600,17 @@ pub fn execute_update_config(
                 .unwrap_or(DEFAULT_AGENTS_EJECT_THRESHOLD),
             min_active_agent_count: min_active_agent_count
                 .unwrap_or(DEFAULT_MIN_ACTIVE_AGENT_COUNT),
+            public_registration: public_registration.unwrap_or(DEFAULT_PUBLIC_REGISTRATION_ENABLED),
```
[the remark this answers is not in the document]
mikedotexe: You're totally right. Added a new error and logic, and unwrapping or using the config's current value. good eye thanks!
fixed in: https://github.com/CronCats/cw-croncat/pull/374/commits/5db464de14b40934dfed59ddc890f62ddfbb8e2f

contracts/croncat-agents/src/tests/agent_tests.rs
mikedotexe: added comment on the Execute method here: https://github.com/CronCats/cw-croncat/pull/374/commits/e368cb006d57ca54701274c0ada2b868d7bc141c

contracts/croncat-agents/src/tests/agent_tests.rs
mikedotexe: Added in a recent commit from a previous comment: https://github.com/CronCats/cw-croncat/pull/374/commits/5db464de14b40934dfed59ddc890f62ddfbb8e2f

contracts/croncat-agents/src/tests/agent_tests.rs
```diff
@@ -2173,10 +2170,46 @@ fn agent_registration_whitelist() {
         .unwrap();
     assert_eq!(
         current_allowed_agents,
-        GetApprovedAgentAddresses {
+        ApprovedAgentAddresses {
             approved_addresses: vec![],
         }
     );
+
+    // Check failure wif admin attempts to "reverse" the public registration from true » false
```
TrevorJTClarke, line 2178: thank you!

TrevorJTClarke (approved):

PR #376, opened 2023-03-06
bug fix addr len check in pause_admin

TrevorJTClarke:


scripts/deployment/src/agents.ts
```diff
@@ -35,9 +43,17 @@ export class AgentClient {
                     "changelog_url": "https://github.com/croncats",
                     "schema": "",
                     "msg": Buffer.from(JSON.stringify({
+                        "pause_admin": `${pauserAddress}`,
```
mikedotexe, line 46: sweet! thank you for sweeping through these TypeScript files!

scripts/deployment/src/agents.ts
```diff
@@ -35,9 +43,17 @@ export class AgentClient {
...
                         "version": `${version[0]}.${version[1]}`,
+                        "public_registration": false,
+                        "allowed_agents": agentsAddresses,
                         "croncat_manager_key": ["manager", version || [0, 1]],
                         "croncat_tasks_key": ["tasks", version || [0, 1]],
+                        // agent_nomination_duration: '',
+                        // min_tasks_per_agent: '',
+                        // min_coins_for_agent_registration: '',
+                        // agents_eject_threshold: '',
```
mikedotexe, line 55: flagging this block of comments in case it's unintentional. I'm happy to leave some cruft in here as we flesh it out tho

TrevorJTClarke: i added this so we know whats available in future, if we ever wanna test it

scripts/deployment/src/agents.ts
TrevorJTClarke: yeah! almost done with the e2e script as well

mikedotexe (approved):
I think this is quite reasonable. Still odd that Bech32 could have varying lengths, but from our chats it sounds like you've dug into this.

I've reached out on the CosmWasm Discord to see if there's additional advice, but this seems like a solid approach right now, albeit a slight head-scratcher when it comes to 63/64

scripts/deployment/src/factory.ts
```diff
@@ -73,4 +73,54 @@ export class FactoryClient {
...
+    contractAddr: string,
+    sub_msg: any,
+    gas: StdFee,
+    funds: any,
+  ): Promise<ExecuteResult> {
+    const proxy_msg = {
+      proxy: {
+        msg: {
+          execute: {
+            contract_addr: contractAddr || '',
```
mikedotexe, line 89: Kinda curious about this line. Feels like we want to make `contract_addr` mandatory to fill out

mikedotexe (approved):
Made some comments, but nothing seems important enough to obstruct forward-movement. Thanks!

scripts/deployment/src/factory.ts
TrevorJTClarke: just me bein dumb :P

PR #377, opened 2023-03-07
added whitelist agent script

TrevorJTClarke:


scripts/deployment/src/whitelist.ts
```diff
@@ -0,0 +1,100 @@
...
+import * as fs from "fs"
+import * as util from "util"
+import { FactoryClient } from './factory';
+config({ path: '.env' })
+// Get values from the environment variables located in the .env file
+const seedPhrase: string = process.env.SEED_PHRASE
+const prefix: string = process.env.PREFIX
+const endpoint: string = process.env.RPC_ENDPOINT
+// NOTE: MUST Be a contract wallet - multisig prefered!
+// If you need one, go to https://github.com/CosmWasm/cw-plus/tree/main/contracts/cw3-fixed-multisig, compile, instantiate & get deployed address.
```
mikedotexe, line 18: love the helpful comments 🚀

scripts/deployment/README.md
```diff
@@ -28,3 +28,10 @@ npm run go
 # runs full scope of contexts for end to end testing
 npm run e2e
 ```
+
+## Add agent to address whitelist
+
+```bash
+# runs full scope of contexts for end to end testing
```
mikedotexe, line 35:
```suggestion
# adds an agent address to the whitelist, so they can register
```

mikedotexe (approved):
So much easier. I'm fine with the postinstall behavior that might overwrite your `.env` file here, as this is just used internally. This will only happen if we add new packages from here on out

PR #378, opened 2023-03-07
Remove vector of hashes

Buckram123:


Buckram123:
Revived #339
Note: checksum not proper

PR #379, opened 2023-03-07
Queries transforms fallback

Buckram123:


Buckram123:
Revived https://github.com/CronCats/cw-croncat/pull/346
Note: checksum not proper

PR #380, opened 2023-03-09
remove updating factory address in tasks contract

TrevorJTClarke:


mikedotexe (approved):
Oh yeah, that is fairly straightforward, isn't it. Man you're fast!

PR #385, opened 2023-03-21
Ft/task query helpers

TrevorJTClarke:
Key updates here:
1. Mod Generic got a new Batch Query handler - this allows for consistent query results to be replicated via RPC for how the internal croncat querier will process queries. Useful for agents. See: https://github.com/CronCats/cw-croncat/blob/ft/task-query-helpers/contracts/mod-generic/src/contract.rs#L79-L181
2. Manager got the new WasmQuery types supported - ALL of them - so now depending on which query you use, it will selectively execute and return each results. NOTE: CodeId query is as of yet untested, as there is not full coverage in the testing libs. See: https://github.com/CronCats/cw-croncat/blob/ft/task-query-helpers/contracts/croncat-manager/src/helpers.rs#L456-L545
3. Task hash & version got added to attributes where necessary
4. Task creation got a more rich response payload, see: https://github.com/CronCats/cw-croncat/blob/ft/task-query-helpers/packages/croncat-sdk-tasks/src/types.rs#L354
5. Changed the definition of `task.is_evented()` because, I should be able to have tasks that evaluate on a block interval within a boundary and have queries. (this was impossible before). So now the definition of "evented" is simply: "Do i have queries AND am i ONCE or IMMEDIATE"? NOTE: i removed the check for queries with "check_result" because there are times like pagination i dont wanna check anything, just do a ton of queries immediately, back to back


contracts/mod-generic/Cargo.toml
```diff
@@ -28,7 +28,7 @@ backtraces = ["cosmwasm-std/backtraces"]
 library = []
 
 [dependencies]
-cosmwasm-std = { workspace = true }
+cosmwasm-std = { workspace = true, features = ["cosmwasm_1_2"] }
```
mikedotexe, line 31: Hmm, that's a new one for me

contracts/mod-generic/src/contract.rs
```diff
@@ -68,3 +69,82 @@ fn generic_query(deps: Deps, query: GenericQuery) -> StdResult<QueryResponse> {
...
+                            Binary::default()
+                        };
+                        // conform response to expected end result,
+                        // always true since we just wanna use values in transforms later
+                        last_response = Some(QueryResponse {
+                            result: true,
+                            data,
+                        });
+                    },
+                    WasmQuery::ContractInfo { contract_addr } => {
```
mikedotexe, line 127: Sweet, this is the same message that allows us to get the code ID and stuff, love having it here

TrevorJTClarke: figured it would allow security if-checks for checksums and code_id changes

contracts/mod-generic/Cargo.toml
TrevorJTClarke:
```
/// The essential data from wasmd's [CodeInfo]/[CodeInfoResponse].
///
/// `code_hash`/`data_hash` was renamed to `checksum` to follow the CosmWasm
/// convention and naming in `instantiate2_address`.
///
/// [CodeInfo]: https://github.com/CosmWasm/wasmd/blob/v0.30.0/proto/cosmwasm/wasm/v1/types.proto#L62-L72
/// [CodeInfoResponse]: https://github.com/CosmWasm/wasmd/blob/v0.30.0/proto/cosmwasm/wasm/v1/query.proto#L184-L199
#[non_exhaustive]
#[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq, JsonSchema)]
#[cfg(feature = "cosmwasm_1_2")]
pub struct CodeInfoResponse {
    pub code_id: u64,
    /// The address that initially stored the code
    pub creator: String,
    /// The hash of the Wasm blob
    pub checksum: HexBinary,
}
```

contracts/mod-generic/src/contract.rs
```diff
@@ -68,3 +70,112 @@ fn generic_query(deps: Deps, query: GenericQuery) -> StdResult<QueryResponse> {
...
+                                contract_addr: contract_addr.clone().to_string(),
+                                msg: msg.clone(),
+                            }
+                            .into(),
+                        );
+                        // conform response to expected end result,
+                        // always true since we just wanna use values in transforms later
+                        match data {
+                            Err(..) => {
+                                last_response = None;
```
mikedotexe, line 113: do we want a `break;` after this, like in the `CosmosQuery::Croncat` variant?

TrevorJTClarke: i would love a break; you offering snacks too? 😹

TrevorJTClarke: and no - we can sandwich non-"checked_result" query that would just be used for getting data inside croncat, here we just want to evaluate it and just pass over it.
Unless you see a use-case where that would be helpful?

contracts/mod-generic/src/contract.rs
```diff
@@ -68,3 +70,112 @@ fn generic_query(deps: Deps, query: GenericQuery) -> StdResult<QueryResponse> {
...
+                            }
+                        }
+                    }
+                    WasmQuery::ContractInfo { contract_addr } => {
+                        let res = deps
+                            .querier
+                            .query_wasm_contract_info(contract_addr.clone().to_string())?;
+                        // conform response to expected end result,
+                        // always true since we just wanna use values in transforms later
+                        last_response = Some(QueryResponse {
```
mikedotexe, line 155: When I try to get the contract info from the CLI and provide an EOA, it does error. (As opposed to returning and empty response or `code_id: ""` or something)

Feels like we might need to check is `res.is_ok()` or `.is_err()` and set `last_response = None` and break?

Trying to make sure someone's accidental inclusion of an EOA gets caught and stops things.

`junod q wasm contract juno1yhqft6d2msmzpugdjtawsgdlwvgq3samrm5wrw`

Screen Shot 2023-03-23 at 9 43 09 AM

TrevorJTClarke: Good catch - resolved in latest push

contracts/croncat-factory/schema/croncat-factory.json
```diff
@@ -387,6 +387,60 @@
             },
             "additionalProperties": false
           },
+          {
+            "description": "Instantiates a new contracts from previously uploaded Wasm code using a predictable address derivation algorithm implemented in [`cosmwasm_std::instantiate2_address`].\n\nThis is translated to a [MsgInstantiateContract2](https://github.com/CosmWasm/wasmd/blob/v0.29.2/proto/cosmwasm/wasm/v1/tx.proto#L73-L96). `sender` is automatically filled with the current contract's address. `fix_msg` is automatically set to false.",
```
mikedotexe, line 391:
```suggestion
            "description": "Instantiates a new contract from a previously uploaded Wasm code using a predictable address derivation algorithm implemented in [`cosmwasm_std::instantiate2_address`].\n\nThis is translated to a [MsgInstantiateContract2](https://github.com/CosmWasm/wasmd/blob/v0.29.2/proto/cosmwasm/wasm/v1/tx.proto#L73-L96). `sender` is automatically filled with the current contract's address. `fix_msg` is automatically set to false.",
```

TrevorJTClarke: @mikedotexe you need to change at the source - this is generated

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -263,7 +263,9 @@ impl Task {
     }
 
     pub fn is_evented(&self) -> bool {
-        self.queries.iter().any(|q| q.check_result)
+        // self.queries.iter().any(|q| q.check_result)
```
mikedotexe, line 266:
```suggestion
```

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -330,14 +342,22 @@ pub struct TaskInfo {
...
+    pub queries: Option<Vec<CosmosQuery>>,
     pub transforms: Vec<Transform>,
     pub version: String,
 }
 #[cw_serde]
 pub struct TaskResponse {
     pub task: Option<TaskInfo>,
 }
+#[cw_serde]
+pub struct TaskExecutionInfo {
```
mikedotexe, line 353: 🤤 thank u

contracts/croncat-manager/src/contract.rs
```diff
@@ -403,7 +396,7 @@ fn end_task(
         amount: coins_transfer,
     };
     Ok(Response::new()
-        .add_attribute("action", "remove_task")
+        .add_attribute("action", "end_task")
```
mikedotexe, line 399: looks good. checked I don't see Vectis relying on the old name 👍🏼

TrevorJTClarke: phew! this def was mis-labeled

contracts/croncat-manager/src/helpers.rs
```diff
@@ -465,22 +579,23 @@ pub fn replace_values(
...
-            serde_json_wasm::to_vec(&action_value)
-                .map_err(|e| StdError::generic_err(e.to_string()))?,
-        );
+        let query_bin = query_response_data
+            .get(transform.query_idx as usize)
+            .ok_or::<Option<Binary>>(None)
+            .unwrap();
+
+        if let Some(bin) = query_bin {
+            let mut q_val = { cosmwasm_std::from_binary(bin)? };
```
mikedotexe, line 590: are these brackets meant to invoke the `.drop` impl? Curious if we should check more rigorously?

not strongly held

contracts/croncat-manager/src/tests/manager_tests.rs
```diff
@@ -4712,6 +4738,158 @@ fn event_task_with_failed_check_result() {
...
+            task_hash: Some(task_hash),
+        },
+        &[], // Attach no funds ofc
+    );
+    assert!(
+        pc_result.is_ok(),
+        "Proxy call should succeed since the check_return comes back true"
+    );
+    let pc_res: AppResponse = pc_result.unwrap();
+    // we expect the queries to pass and the task to reschedule after success
```
mikedotexe, line 4865: interesting, I thought we would have to fast-forward block/time in order to see the reschedule happen

TrevorJTClarke: its the response from proxy_call, no need to fwd time

contracts/croncat-manager/src/helpers.rs
TrevorJTClarke: no, that was leftovers... removing

PR #389, opened 2023-03-27
updated agent info to return stats

### LINKED ISSUE #388: Agent stats missing from getter
See: https://github.com/CronCats/cw-croncat/blob/main/contracts/cw-croncat/src/agent.rs#L44

During the upgrade to new architecture, this amount was missed.
Since we already have stats stored elsewhere, just needs simple response addition.

TrevorJTClarke:
closes #388


mikedotexe (approved):
awesome, will be great to have access to that info

PR #390, opened 2023-03-27
Use correct `Value` reference

mikedotexe:
As discussed on Friday, we were accidentally using the `Value`.


TrevorJTClarke (approved):
Muchas Garcias

PR #391, opened 2023-03-27
Include `TransactionInfo` in `TaskExecutionInfo`

mikedotexe:
The purpose of this PR is to provide a new state variable that acts as a simple register, saving limited but useful details on what task was just executed. This state variable is saved with default values on instantiation, and then each time `proxy_call` passes through all guards, it saves this info right before the submessages are sent.


packages/croncat-sdk-manager/src/types.rs
```diff
@@ -5,6 +5,8 @@ use cw20::Cw20CoinVerified;
 
 use crate::error::SdkError;
 
+pub const LAST_TASK_EXECUTION_INFO_KEY: &str = "last_task_execution_info";
```
mikedotexe, line 8: After talking with Vectis folks, it feels like it makes more sense to have the state keys saved on the packages level, so integrators can have more confidence that they're not hardcoding strings and hoping they don't change.

Just a nice-to-have, but perhaps we can keep this in mind if we add new state variables in future versions.

TrevorJTClarke: This seems fine for this type of scenario. Doesnt 100% make sense for all.

TrevorJTClarke (approved):
looks great!

PR #397, opened 2023-04-07
Changed block logic to move current block minus 1

TrevorJTClarke:


contracts/croncat-tasks/src/contract.rs
```diff
@@ -661,7 +670,7 @@ fn query_current_task(deps: Deps, env: Env) -> StdResult<TaskResponse> {
             .range(
                 deps.storage,
                 None,
-                Some(Bound::inclusive(env.block.time.nanos())),
+                Some(Bound::inclusive(env.block.time.plus_nanos(6).nanos())),
```
mikedotexe, line 673: What's this for?

mikedotexe (approved):

contracts/croncat-tasks/src/contract.rs
TrevorJTClarke: to keep the block & timestamp scope within the same margin. I know we talked about not bothering with timestamps, but i wanted to be consistent. The margin of error here will be lower than not doing this, esp based on real world latencies, so ive added.

PR #399, opened 2023-04-07
End boundary now respects the end block without extra remove execution

TrevorJTClarke:


packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -486,7 +486,11 @@ pub(crate) fn get_next_block_by_offset(
                 end.u64()
             };
             // we ONLY want to catch if we're passed the end block height
-            std::cmp::min(next_block_height, end_height)
+            if next_block_height > end_height {
```
mikedotexe, line 489: I think I see now. A few lines above this we have:

https://github.com/CronCats/cw-croncat/pull/399/files#diff-bd67bce8e428efb0ad94009abe2def6c51aff31014b83a0df1c69531246485feR479

but that's only saying, "is now after the boundary end?" but we allowed for rescheduling past the end boundary. Makes much more sense to see it

mikedotexe (approved):
tysm for hunting this down 𐃆

packages/croncat-sdk-tasks/src/types.rs
TrevorJTClarke: yep exactly - we missed this spot :/

PR #400, opened 2023-04-10
PathToValue update, E2E tasks minor tweak to schema

TrevorJTClarke:
Updates:
1. Fixed e2e evented tasks schema
2. Fixed consistency in mod-balances response data to use Coin
3. Fixed mod-balances balance comparison tests
4. Added full test coverage for PathToValue
5. Enabled empty pointer in PathToValue so that values returned in query can be simple non-map or non-vec string values. Many contracts actually just respond with a string value like Uint64 "123456789", so without an empty pointer, we couldn't support this.


contracts/mod-balances/src/contract.rs
```diff
@@ -141,7 +141,11 @@ fn query_has_balance_comparator(
                 required_cw20.address.clone(),
                 &cw20::Cw20QueryMsg::Balance { address },
             )?;
-            (balance_response.balance, required_cw20.amount)
+            (
+                balance_response.balance,
+                required_cw20.address.to_string(),
```
mikedotexe, line 146: 👍🏼

contracts/mod-generic/src/tests/path_to_value.rs
```diff
@@ -0,0 +1,239 @@
...
+    let mut value = Map(BTreeMap::new());
+    let path = PathToValue(vec![]);
+    assert_eq!(path.find_value(&mut value.clone()).unwrap(), &mut value);
+}
+
+#[test]
+fn test_find_value_empty_path_realworld_case() {
+    let mut value = Value::String("raw_string_returned".to_string());
+    let path = PathToValue(vec![]);
+    assert_eq!(path.find_value(&mut value.clone()).unwrap(), &mut value);
```
mikedotexe, line 19: I feel like I'm not understanding something simple here.

So we set `path` to be `PathValue` initialized with an empty vector, but then we're trying to find a value. But when did we add the value to `path`?

TrevorJTClarke: We're creating an empty path, which points at the root of the response value. 
Imagine "let mut value" here, is coming back from a query, and "let path" is the pointer to extract data from query. In this realworld_case ™️ we are only getting a string back that isn't a valid object or array, so we just want path: [] to point at the root.

mikedotexe (approved):
I feel like I'm missing some internals about `PathToValue` but it's not concerning me much.

PR #402, opened 2023-04-12
Add integration packages

mikedotexe:
New packages for integrating contracts, providing CronCat workflow helpers, a proc macro, and test helpers.

## Task creation

- `get_latest_croncat_contract` — provided the CronCat factory address and a contract name (eg. "tasks") it'll return the address of the latest contract, or throw a custom error.
- `create_croncat_task_submessage` — this is likely a function that integrating dApps will use frequently. DApps are able to create a `TaskRequest` object (exported as `CronCatTaskRequest`), provide the factory address and optional reply parameters. It will return a Result with a submessage that can be added to the dApps' Response, like `Response::new().add_submessage(s)(…`
- `create_croncat_task_message` — same as the submessage one above, except as a message, for integrating dApps that have decided they don't need a reply, for whatever reason they've landed on.
- `create_croncat_task_cosmos_msg` — this function is being used as a helper during the previous calls, but might as well be exported for developers, as doing so can only be helpful. It returns a Result of the Wasm Execute message to create the CronCat task.
- `croncat_basic_validation` — this function can grow in the future if we wish, and only contains one validation, which is ensuring that funds have been attached. Due to the nature of error handling, it feels like we must attempt to do basic validation in order to reduce DevX confusion from integrators.
After talking to an integrating partner, ensured there are two structs to aid in flexibility.

```rs
pub struct CronCatTaskSubmessageParams {
    /// Defaults to [REPLY_CRONCAT_TASK_CREATION](crate::REPLY_CRONCAT_TASK_CREATION)
    pub reply_id: Option<u64>,
    /// Defaults to [Always](SubMessageReplyType::Always)
    pub reply_type: Option<SubMessageReplyType>,
}
```

The `REPLY_CRONCAT_TASK_CREATION` is set to a large, unique number derived from the word "croncat" so as to reduce collisions in integrating dApps.

and a simple one for Reply type, which I didn't find elsewhere:

```rs
pub enum SubMessageReplyType {
    Always,
    OnError,
    OnSuccess,
}
```

## Handle reply after task creation

When you create a task using a submessage with replies, you can utilize two functions in your contract's reply entry point.

- `reply_handle_task_creation` — this takes the `msg: Reply` argument and parses it into the more useful `CronCatTaskExecutionInfo`, which is just the `croncat-sdk-tasks`'s `TaskExecutionInfo`. It returns a tuple with the logic in the reply entry point and the binary message data. The latter is useful to include in the reply's Response, like by using `Reply::new().set_data(…`.

## Handle invocation from a task

Method introduced here:
- `handle_incoming_task` — this will check the version of the task, and confirm (via a factory query) that the sender is a sanctioned manager contract. It also checks that the execution is happening synchronously by confirming via state that the manager saves right before sending the messages invoking it. Finally, it checks checks that the owner of the task is the receiving contract itself, or else the `expected_owner` provided in the optional parameters.

In this scenario, your contract is getting called, at a method that we expect to be called by CronCat. We want to know if indeed it's CronCat executing your task, and also check that the task's owner is the receiving contract. If you wish to modify these restrictions, you are able to with this struct for additional parameters:

```rs
pub struct HandleIncomingTaskParams {
    /// Disables the check ensuring that the block height and transaction index are the same
    /// If you expect an IBC delay or something asynchronous, disable by setting to true.
    pub disable_sync_check: bool,
    /// By default, we check that the contract receiving the task invocation
    /// must be the owner of the that task. Put another way: someone else's
    /// task isn't invoking our method. You can disable this by setting to true.
    pub disable_owner_check: bool,
    /// If the owner check is enabled, you may specify an alternate expected owner.
    /// Perhaps the task owner isn't this contract, but you know the address.
    /// By default, the validation logic in `handle_incoming_task` checks against the current contract.
    /// If disable_owner_check is true, this value is irrelevant.
    pub expected_owner: Option<Addr>,
}
```

------

# Testing helpers

The `croncat-integration-testing` package has a few helper constants, functions, and `cw-multi-test` exports that'll aid dApps in testing their custom workflows that leverage CronCat automation.

- `set_up_croncat_contracts` — you may provide an optional `cw-multi-test` `App` object, and it will set up the CronCat contracts and return this struct with helpful variables to be used in unit tests.

```rs
pub struct CronCatTestEnv {
    pub app: cw_multi_test::App,
    pub factory: cosmwasm_std::Addr,
    pub manager: cosmwasm_std::Addr,
    pub tasks: cosmwasm_std::Addr,
    pub agents: cosmwasm_std::Addr,
}
```

- `add_seconds_to_block` — convenience method to move time forward in the `cw-multi-test` environment
- `increment_block_height` — convenience method to increase block height in the `cw-multi-test` environment 

There are also other exposed methods that will allow integrators to store and instantiate the various CronCat contracts.

------

## Error macro

This is a procedural attribute macro that creates `#[croncat_error]` that is intended to be placed above the integrating contract's error enum, which is conventionally called `ContractError`.

**Note**: this macro should go above the usual `Derive` we see above this enum. The correct placement looks like this:

```rs
#[croncat_error]
#[derive(Error, Debug, PartialEq)]
pub enum ContractError {
  // variants here…
}
```

It will throw a compile-time error if it is not placed correctly.

The macro adds this enum variant:

```rs
#[error("CronCat error: {err:?}")]
  CronCatError {
  err: CronCatContractError
}
```

It also adds this logic:

```rs
impl From<CronCatContractError> for ContractError {
  fn from(error: CronCatContractError) -> Self {
    ContractError::CronCatError {
      err: error,
    }
  }
}
```

The above logic will allow helper functions in `croncat-integration-utils` to propagate errors in a manner consistent with the contract's errors.

------

## Justfile

Some of these new packages cannot be compiled to the target `wasm32-unknown-unknown`, and so the build command has been modified.


packages/croncat-integration-utils/README.md
```diff
@@ -0,0 +1,3 @@
+# CronCat Integration Utilities
+
+This repo contains helper methods for projects adding automation to their smart contracts using CronCat tasks.
```
TrevorJTClarke, line 3: Can you add the PR comment into this readme? That explainer was great

packages/croncat-integration-utils/src/reply_handler.rs
```diff
@@ -0,0 +1,45 @@
+use crate::error::CronCatContractError;
+use crate::{CronCatTaskExecutionInfo, REPLY_CRONCAT_TASK_CREATION};
+use cosmwasm_std::{Reply, Response, Uint64};
+use cw_utils::parse_reply_execute_data;
+
+/// Reply handler when a contract calls [`create_task`](croncat_sdk_tasks::msg::TasksExecuteMsg::CreateTask).
+/// This will handle [`reply_always`](cosmwasm_std::ReplyOn::Always) covering success and failure.
+pub fn reply_handle_task_creation(
```
TrevorJTClarke, line 8: nice! I think it may be better to explicitly call this: `reply_handle_croncat_task_creation`

packages/croncat-integration-utils/src/reply_handler.rs
```diff
@@ -0,0 +1,45 @@
...
+    // We return the newly-created task details
+    // in your contract's state if you wish.
+    // Please see the create-task-handle-tick example for info:
+    // https://github.com/CronCats/cw-purrbox/tree/main/contracts
+    Ok(created_task_info)
+}
+
+/// This helps return a format of the latest [task execution info](croncat_sdk_tasks::types::TaskExecutionInfo).
+/// Return a JSON representation in the Response's data.
+pub fn reply_complete_task_creation(
```
TrevorJTClarke, line 39: Not entirely following why this is needed if the types are available, they can handle how they need.
Maybe better use case somewhere?
Another explicitly named: `reply_data_croncat_task_creation`

packages/croncat-integration-utils/src/handle_incoming_task.rs
```diff
@@ -0,0 +1,119 @@
...
+use croncat_sdk_tasks::types::TaskExecutionInfo;
+
+/// Handles and validates an incoming CronCat task
+/// Specifically, it checks:
+/// - Sender is a sanctioned CronCat manager contract (CronCat factory knows the manager contract addresses and versions.)
+/// - We're in the same block and transaction index as the latest executed transaction. In other words, this task is happening in an atomic, synchronous transaction, and this invocation is the result of a cross-contract call (message or submessage) from a manager contract. (Note: you can disable this check via the `disable_sync_check` field of `custom_validation` if, for instance, you're doing IBC calls where the execution is asynchronous, spanning blocks.)
+/// - The owner of the task that just called is the calling contract (Note: this can be changed by setting the `expected_owner` field in `custom_validation`. If unset, it will default to this contract.)
+/// For contracts storing task hashes, you can take the [`TaskExecutionInfo`](croncat_sdk_tasks::types::TaskExecutionInfo) returned and check it against state.
+pub fn handle_incoming_task(
+    querier: &QuerierWrapper,
```
TrevorJTClarke, line 15: curious why you chose this over deps?

packages/croncat-integration-utils/src/handle_incoming_task.rs
```diff
@@ -0,0 +1,119 @@
...
+    // Catching deserialization issues
+    if latest_task_execution_cast_res.is_err() {
+        return Err(CronCatContractError::DeserializeTaskInfo {});
+    }
+
+    // Pertinent info containing, among other things, the task version
+    let latest_task_execution: TaskExecutionInfo = latest_task_execution_cast_res.unwrap();
+
+    // We turn (for example) "0.1" into [0, 1] so we can query the factory with this value and the contract name ("manager")
+    let versions = latest_task_execution
```
TrevorJTClarke, line 59: nice - yeah still been thinking on why we ever did string... lol someday that can change

packages/croncat-integration-utils/src/handle_incoming_task.rs
```diff
@@ -0,0 +1,119 @@
...
+        .version
+        .split('.')
+        .map(|v| -> u8 { v.parse().unwrap() })
+        .collect::<Vec<u8>>();
+
+    let mut state_key =
```
TrevorJTClarke, line 65: this is great!

packages/croncat-integration-utils/src/handle_incoming_task.rs
```diff
@@ -0,0 +1,119 @@
...
+    if sanctioned_manager_res.is_none() {
+        return Err(CronCatContractError::FactoryManagerQueryFailed {
+            manager_addr: sender,
+            version: latest_task_execution.version,
+        });
+    }
+
+    let sanctioned_manager_address = sanctioned_manager_res.unwrap();
+
+    let quoted_sender = format!(r#""{}""#, sender);
```
TrevorJTClarke, line 81: interesting - is that because of how raw escapes things?

packages/croncat-integration-utils/src/handle_incoming_task.rs
```diff
@@ -0,0 +1,119 @@
...
+        return Err(CronCatContractError::UnsanctionedInvocation {
+            manager_addr: sender,
+            version: latest_task_execution.version,
+        });
+    }
+
+    // If this method is called normally (with disable_sync_check defaulting to false)
+    // This will check for synchronous invocation from the CronCat manager.
+    // This method can be called, ignoring this check by setting it to `true`.
+    if !disable_sync_check {
```
TrevorJTClarke, line 96: egggselllent

packages/croncat-integration-utils/src/handle_incoming_task.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        if !is_same_block_bool || !is_same_tx_id_bool {
+            return Err(CronCatContractError::NotSameBlockTxIndex {});
+        }
+    }
+
+    // Last, we check if the task creator is this contract, ensuring
+    // this invocation hasn't happened from someone else's task.
+    // In cases where that's too restrictive, you may specify
+    if !disable_owner_check && latest_task_execution.owner_addr != owner {
```
TrevorJTClarke, line 112: w00t

TrevorJTClarke (changes requested):
great work man!!
Left a couple minor changes -- but lots of solid progress!

Cargo.toml
```diff
@@ -76,4 +76,5 @@ croncat-sdk-tasks = { path = "./packages/croncat-sdk-tasks" }
 mod-sdk = { path = "./packages/mod-sdk" }
 croncat-sdk-factory = { path = "./packages/croncat-sdk-factory" }
 croncat-sdk-manager = { path = "./packages/croncat-sdk-manager" }
-cw-boolean-contract = "0.1.2"
\ No newline at end of file
+cw-boolean-contract = "0.1.2"
+cosmwasm-storage = "1.2.3"
```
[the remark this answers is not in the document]
mikedotexe: Good eye, as mentioned offline, have to remove cosmwasm-storage, and have borrowed a couple functions instead

packages/croncat-integration-utils/src/handle_incoming_task.rs
mikedotexe: yeah when you deserialize/decode it comes back as `"value"` instead of `value` so wrap it in quotes, and yeah need double-double quotes, animal-style.

justfile
```diff
@@ -54,8 +68,7 @@ optimize:
 		--mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
 		--mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
 		--platform linux/amd64 \
-		cosmwasm/workspace-optimizer:0.12.11
-
+		cosmwasm/workspace-optimizer:0.12.12
```
TrevorJTClarke, line 71: careful here - this tiny version bump has controversy on OS support

packages/croncat-integration-utils/README.md
mikedotexe: done! did this for the other packages, too.

for some reason I can tease the writing muse when making a pull request description easier than sitting and writing the README. Maybe I'll use this tactic going forward. I hope myself doesn't find out about it.

packages/croncat-integration-utils/src/reply_handler.rs
mikedotexe: Removed this in favor of returning a tuple from `reply_handle_croncat_task_creation`

packages/croncat-integration-utils/src/handle_incoming_task.rs
mikedotexe: We don't use the storage, and think people will familiar with CronCat might be more at ease knowing that we're not touching their storage

packages/croncat-integration-utils/src/handle_incoming_task.rs
mikedotexe: 😇 
🚀

packages/croncat-integration-utils/src/handle_incoming_task.rs
mikedotexe: So I took Vectis's advice and made this much simpler, no longer need this and am doing it from storage more directly.

justfile
mikedotexe: good eye, update to workspace-optimizer:0.12.13 here: https://github.com/CronCats/cw-croncat/pull/402/commits/95bf365469b3e728e5ab80354dc6ee38a51cea26

justfile
```diff
@@ -54,8 +68,7 @@ optimize:
 		--mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
 		--mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
 		--platform linux/amd64 \
-		cosmwasm/workspace-optimizer:0.12.11
-
+		cosmwasm/workspace-optimizer:0.12.13
```
TrevorJTClarke, line 71: can confirm 0.12.13 works on my OS version. phew!

TrevorJTClarke (approved):
huge home run sir. TYSM!
well doxxed, well thought out setup.

.github/workflows/Basic.yml
```diff
@@ -29,12 +28,21 @@ jobs:
...
-          command: wasm
-          args: --locked
-        env:
-          RUSTFLAGS: "-C link-arg=-s"
+        run: |
+          set -e
+          export RUSTFLAGS='-C link-arg=-s'
+          EXCLUDED_PACKAGE="croncat_integration_testing"
+          
+          for PACKAGE in $(cargo metadata --format-version 1 --no-deps | jq '.packages[] | .name' -r)
```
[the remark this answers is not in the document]
mikedotexe: I have a friend named Gary Petunia Thielson that helped me

mikedotexe:
Made exact version so we don't accidentally use cosmwasm-std 1.2.x since that's not deployed to many blockchains.

The lockfile has an update that's 1.2.x but it's for a proc macro:
https://github.com/CosmWasm/cosmwasm/blob/v1.2.4/packages/derive/Cargo.toml

TrevorJTClarke:
Where are we at with this? Are you hoping to merge and release? or wait until something?

integration-sdk/croncat-errors-macro/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+use syn::{parse_macro_input, Data, DataEnum, DeriveInput, Error, Ident, Variant};
+
+#[proc_macro_attribute]
+pub fn croncat_error(_attrs: TokenStream, input: TokenStream) -> TokenStream {
+    let mut input = parse_macro_input!(input as DeriveInput);
+    let enum_name = input.ident.clone();
+
+    // Ensure they've placed the attribute macro above the derive
+    if !input.attrs.iter().any(|attr| attr.path.is_ident("derive")) {
+        let msg = "😻 Please move the `#[croncat_error]` macro above the `#[derive]` attribute with the `Error` trait. 😻";
```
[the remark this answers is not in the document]
mikedotexe: Correct, when they go to compile their contract it'll say, "Couldn't compile because of an error" and they scroll up and see the cute cat emojis that tell them exactly what to do.

integration-sdk/croncat-integration-utils/src/reply_handler.rs
```diff
@@ -0,0 +1,35 @@
+use crate::error::CronCatContractError;
+use crate::{CronCatTaskExecutionInfo, REPLY_CRONCAT_TASK_CREATION};
+use cosmwasm_std::{Binary, Reply, Uint64};
+use cw_utils::parse_reply_execute_data;
+
+/// Reply handler when a contract calls [`create_task`](croncat_sdk_tasks::msg::TasksExecuteMsg::CreateTask).
+/// This will handle [`reply_always`](cosmwasm_std::ReplyOn::Always) covering success and failure.
+pub fn reply_handle_croncat_task_creation(
```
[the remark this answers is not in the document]
mikedotexe: I know I've had a few moments where I'm like, "dang it, I have the bytes, how the hell do I turn this into my struct again? oh yeahhh, that `serde_json_wasm`, forgot about that.

I also rarely use `cw_utils` and it doesn't exist in any of the examples.

Basically making things easy, zero cognitive overhead for projects that wish to use CronCat

integration-sdk/croncat-integration-utils/src/task_creation.rs
```diff
@@ -0,0 +1,119 @@
...
+/// See [TASKS_NAME](crate::TASKS_NAME), [MANAGER_NAME](crate::MANAGER_NAME), and [AGENTS_NAME](crate::AGENTS_NAME)
+pub fn get_latest_croncat_contract(
+    querier: &QuerierWrapper,
+    croncat_factory_address: Addr,
+    croncat_contract_name: String,
+) -> Result<Addr, CronCatContractError> {
+    let query_factory_msg = LatestContract {
+        contract_name: croncat_contract_name.clone(),
+    };
+    let latest_contract_res: ContractMetadataResponse =
```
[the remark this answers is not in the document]
mikedotexe: Agree this is good to do but not urgent. Given our offline convo today, would like to march forward, having captured this in this ticket: https://github.com/CronCats/cw-croncat/issues/409

integration-sdk/croncat-integration-utils/src/tests/mod.rs
```diff
@@ -0,0 +1 @@
+
```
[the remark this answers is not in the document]
mikedotexe: I've thought about this a bit as well.

In terms of the CronCat product, I believe the only thing these integration crates are doing is *using* what exists and is merged into the codebase. Example of an existing test for the TaskExecutionInfo check during incoming invocation: https://github.com/CronCats/cw-croncat/blob/beta-0.1.5/contracts/croncat-manager/src/tests/manager_tests.rs#L5361

`cw-multi-test` is slightly limited in certain kinds of robusts tests involving multiple transactions in a block, so I added this issue after Simon referred me here from another issue I created: https://github.com/CosmWasm/cw-multi-test/issues/36

Mentally, I came back to the topic of test coverage for this, and couldn't convince myself of its importance.

Here's an example from the (slightly more complicated) example "create a tick task to myself and make sure no one else can call me at this method" here, showing the full flow of deploying an integrating dApp and handling the reply correctly: https://github.com/CronCats/cw-purrbox/blob/main/contracts/create-task-handle-tick/src/tests/create_task_handle_tick.rs#L104

Here's another test in that same file that adds an agent and proxy call into the mix: https://github.com/CronCats/cw-purrbox/blob/main/contracts/create-task-handle-tick/src/tests/create_task_handle_tick.rs#L247

packages/croncat-sdk-tasks/src/types.rs
```diff
@@ -192,12 +192,12 @@ pub struct Transform {
     /// Action key path to the value that should get replaced
     /// for example:
     /// X: {Y: {Z: value}}
-    /// [X,Y,Z] to reach that value
+    /// \[X,Y,Z\] to reach that value
```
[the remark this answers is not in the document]
mikedotexe: `cargo doc` told me to

PR #410, opened 2023-05-03
Fix kicked attribute

mikedotexe:


TrevorJTClarke (approved):
you have no idea how much better i feelz

PR #414, opened 2023-05-16
intergration-sdk/reply-id

whalelephant:
- the `id` in `Reply` should be the one set by the caller
- added `ParseReplyError` conversion to show error (optional)


integration-sdk/croncat-integration-utils/src/error.rs
```diff
@@ -7,8 +8,11 @@ pub enum CronCatContractError {
     #[error("{0}")]
     Std(#[from] StdError),
 
+    #[error("Parse Reply Error {0}")]
+    ParseReplyError(#[from] ParseReplyError),
```
mikedotexe, line 12: So much better

mikedotexe (approved):

PR #418, opened 2023-05-25
Tasks contract: update `validate_boundary` logic and extend tests

mikedotexe:
Some of the tests were throwing `InvalidBoundary` on something that's possible.

Kudos to Kado for helping uncover this issue


contracts/croncat-tasks/src/helpers.rs
```diff
@@ -29,7 +32,7 @@ pub(crate) fn validate_boundary(
             }
         }
         (
-            Interval::Block(_) | Interval::Once | Interval::Immediate,
+            Interval::Block(_) | Interval::Once | Interval::Immediate | Interval::Cron(_),
```
Buckram123, line 35:
```suggestion
            _,
```
Since it's for any `Interval` now :thinking:. Unless we plan to add other interval later on

contracts/croncat-tasks/src/helpers.rs
```diff
@@ -775,7 +815,137 @@ mod tests {
...
+                    start: Some(Timestamp::from_nanos(123456)),
+                    end: Some(Timestamp::from_nanos(123457)),
+                })),
+                123,
+                Timestamp::from_nanos(123456),
+                Ok(Boundary::Time(BoundaryTime {
+                    start: Some(Timestamp::from_nanos(123456)),
+                    end: Some(Timestamp::from_nanos(123457)),
+                })),
+            ),
```
Buckram123, line 846: Is it fine that we have here 1 nanosecond boundary?

Buckram123 (approved):
Nice to see more flexibility

contracts/croncat-tasks/src/helpers.rs
mikedotexe: It's very silly, for sure. I think for now the main check is that the end is greater than the start.

contracts/croncat-tasks/src/helpers.rs
mikedotexe: Good one. I want to take a closer look at `Interval::Immediate` elsewhere in the code. Need to remember what use case differentiates `Immediate` and `Once` when there's a boundary involved.

Like if I say, "do this task at the boundary start of 5 blocks from now" it feels like both `Immediate` and `Once` will behave similarly. Perhaps the only difference is `Immediate` would technically be recurring, and fire every block after the boundary start and not after boundary end. Need to investigate a bit to confirm.

contracts/croncat-tasks/src/helpers.rs
Buckram123: I think it's within `slot_granularity_time` range, so it's not gonna trap agent https://github.com/CronCats/cw-croncat/blob/b1206fd4d363108e1d829dfd67427f93bffc7ec2/packages/croncat-sdk-tasks/src/types.rs#L39-L40

PR #422, opened 2023-06-13
Ignore integration coverage

Buckram123:


mikedotexe (approved):
thank you for this suggestion and follow-through!

PR #424, opened 2023-06-23
Simulate tasks for integrations

Buckram123:
Simplest simulate without any checks.
Main usage is to calculate amount for one task, etc...


Buckram123:
Wonder, do we need a strict version, thoughts? @mikedotexe

integration-sdk/croncat-integration-utils/src/task_creation.rs
```diff
@@ -151,3 +156,116 @@ pub fn croncat_basic_validation(info: &MessageInfo) -> Result<(), CronCatContrac
...
+        match &action.msg {
+            CosmosMsg::Wasm(WasmMsg::Execute {
+                contract_addr,
+                funds,
+                msg,
+            }) => {
+                for coin in funds {
+                    if coin.amount.is_zero() || !amount_for_one_task.add_coin(coin.clone()).unwrap()
+                    {
+                        // Ignore errors, we assuming we have valid task here
```
mikedotexe, line 211: Not entirely sure I follow here. Is this for people who want to create a task and will run this beforehand to make sure they know the gas needs? If so, feels like we might want to give them an error? Trying to catch up ;)

Buckram123: Hey more like create task for X executions, but they don't know `AmountForOneTask` beforehand. Also wanted to give minimal input as possible. Not sure if we want to expose full checks, since it may be harder for the user to fill those inputs, I'd rather have query on tasks contract for error checks, thoughts?

mikedotexe (approved):
thanks for the patience on the review!

